PR #36, opened 2019-05-02
Update OfferSearchController bootstrapping

bertramakers:


src/Offer/OfferControllerProvider.php
```diff
@@ -42,18 +46,29 @@ public function __construct(
      */
     public function connect(Application $app)
     {
+        $app['offer_search_request_parser'] = $app->share(
```
jdrieghe: Is it necessary to share this on the container? Could we just create the `CompositeOfferRequestParser` instance and put it in a variable here?

PR #38, opened 2019-05-24
Update bootstrapping of offer and organizer controllers

bertramakers:


src/Organizer/OrganizerControllerProvider.php
```diff
@@ -13,11 +15,15 @@ public function connect(Application $app)
...
             function (Application $app) {
+                $requestParser = (new CompositeOrganizerRequestParser())
+                    ->withParser(new WorkflowStatusOrganizerRequestParser());
+
                 return new OrganizerSearchController(
                     $app['organizer_elasticsearch_query_builder'],
                     $app['organizer_elasticsearch_service'],
-                    $app['paged_collection_factory'],
-                    $app['elasticsearch_query_string_factory']
+                    $requestParser,
```
jdrieghe: We're changing constructor arguments here for a Controller but I don't see changes to that Controller. Did we forget anything?

bertramakers: Yes, it's changed here https://github.com/cultuurnet/udb3-search-http/pull/49

bertramakers: Forgot to add you as a reviewer though. In general it's always best to check the PRs linked in the issue though since we usually do QA per ticket and not per PR since there's so many intertwined packages :)
But I still need to test this manually so there might still be bugs.

PR #39, opened 2019-08-12
Internalize udb3-search-http

jdrieghe:
### Changed
- `cultuurnet/udb3-search-http` is now fully encapsulated in `cultuurnet/udb3-search-service`
- Upgraded PHPUnit to version `7.5`
- Moved App related code to `app/` folder

### Fixed
- Code styles now adhere to stricter standard
---
Ticket: https://jira.uitdatabank.be/browse/III-3033


bertramakers:
Maybe we should move everything that's in the `src` folder on `master` to `app` first, like we did with `udb3-silex`? Then we can put the http library code in `src/Http` instead of `src/Search/Http`. The whole app is search so `src/Search` would be a bit redundant anyway.

jdrieghe:
@bert I followed the existing namespace pattern for now. I'll add the move to `app/` in this PR if that's okay?

bertramakers:
The namespace is correct yes, I meant the directory structure. I think we could just map it as `CultuurNet\UDB3\Search` => `src` in composer.json if we move the other code to `app` first?

jdrieghe:
@bertramakers Gotcha. Done!

PR #43, opened 2019-08-23
Index id property of duplicated event location as array of ids

bertramakers:
### Changed

- The `location.id` on event documents is now indexed as an array of ids when the location has known duplicates.

---

Ticket: https://jira.uitdatabank.be/browse/III-3046

---

This does not require a schema change because Elasticsearch has no `array` datatype. Every field can always be indexed as `0...n` values out of the box: https://www.elastic.co/guide/en/elasticsearch/reference/5.2/array.html


src/ElasticSearch/JsonDocument/CopyJson/CopyJsonRelatedLocation.php
```diff
@@ -54,6 +59,7 @@ public function __construct(
         FallbackType $fallbackType
     ) {
         $this->logger = $logger;
+        $this->idUrlParser = $idUrlParser;
```
bertramakers: Note that this dependency was already being injected, but not stored as a property (because it was passed to another object instead).

PR #45, opened 2019-08-28
III-3064 - Remove dependency on cultuurnet/udb3

jdrieghe:
### Added
- Events that were needed to deserialize queued messages
- Value objects that were needed by this service but lived in `cultuurnet/udb3`
- Some composer dependencies that were implicitly loaded through `cultuurnet/udb3`

### Changed
- We now only know and care about a limited subset of events that can be broadcast by `udb3`. Everything else is ignored by a change to `cultuurnet/broadway-amqp` (https://github.com/cultuurnet/broadway-amqp/pull/12)
- Upgraded `satooshi/php-coveralls` to `php-coveralls/php-coveralls` version `^2.1`

### Removed
- Dependency on `cultuurnet/udb3`
---
Ticket: https://jira.uitdatabank.be/browse/III-3064


composer.json
```diff
@@ -12,11 +12,13 @@
...
         "php": "^7.1",
         "2dotstwice/silex-feature-toggles-provider": "~0.1",
+        "broadway/broadway": "~0.10",
+        "cakephp/chronos": "^1.1",
         "cultuurnet/culturefeed-php": "~1.9.2",
+        "cultuurnet/deserializer": "~0.1",
         "cultuurnet/geocoding": "~0.1",
         "cultuurnet/hydra": "~0.1",
         "cultuurnet/silex-amqp": "~0.1",
-        "cultuurnet/udb3": "~0.1",
```
bertramakers:
```
composer install
Package operations: [...] 27 removals
```
🎉

bootstrap.php
```diff
@@ -138,13 +141,13 @@ function () {
...
-        $maps =
-            \CultuurNet\UDB3\Event\Events\ContentTypes::map() +
-            \CultuurNet\UDB3\Place\Events\ContentTypes::map() +
-            \CultuurNet\UDB3\Label\Events\ContentTypes::map() +
-            \CultuurNet\UDB3\Organizer\Events\ContentTypes::map();
+        $deserializerMapping = [
+            EventProjectedToJSONLD::class => 'application/vnd.cultuurnet.udb3-events.event-projected-to-jsonld+json',
+            PlaceProjectedToJSONLD::class => 'application/vnd.cultuurnet.udb3-events.place-projected-to-jsonld+json',
+            OrganizerProjectedToJSONLD::class => 'application/vnd.cultuurnet.udb3-events.organizer-projected-to-jsonld+json',
+        ];
```
bertramakers: I double checked to make sure and these should indeed be the only messages we actually do something with. 👍 
(Checked out the branch locally, ran composer install, and looked at the various implementations of `EventListenerInterface`)

src/JsonDocument/TransformingJsonDocumentIndexService.php
```diff
@@ -33,12 +33,12 @@ class TransformingJsonDocumentIndexService implements
...
      * @param ClientInterface $httpClient
      * @param JsonDocumentTransformerInterface $jsonDocumentTransformer
-     * @param DocumentRepositoryInterface $searchRepository
+     * @param ElasticSearchDocumentRepository $searchRepository
      */
     public function __construct(
         ClientInterface $httpClient,
         JsonDocumentTransformerInterface $jsonDocumentTransformer,
-        DocumentRepositoryInterface $searchRepository
+        ElasticSearchDocumentRepository $searchRepository
```
bertramakers: This class is now tightly coupled to an Elasticsearch implementation. Can we copy over the `DocumentRepositoryInterface` as well so we can keep this decoupled?

jdrieghe: Yes we can. I'll quickly do that so we can merge this beast :D

jdrieghe: Done

PR #47, opened 2019-09-09
III-3004 Remove cache layer

jdrieghe:
### Changed
- Upgraded dependency on `udb3-api-guard` to latest version that no longer includes the cache layer

### Fixed
- Incorrectly passed `boolean` parameter to `ApiKeyRequestAuthenticator`


bertramakers:
I think this controller provider should be removed as well (as well as where it's `mount`ed)? It would now clear the cache in the `InMemoryRepository` but that makes no sense. The controller class also no longer exists. https://github.com/cultuurnet/udb3-search-service/blob/master/app/ApiGuard/ApiGuardControllerProvider.php

Other than that this looks good 👍

app/ApiGuard/ApiGuardServiceProvider.php
```diff
@@ -39,7 +37,7 @@ function (Application $app) {
                 return new CultureFeedApiKeyAuthenticator(
                     $app['culturefeed'],
                     $app['auth.consumer_repository'],
-                    $app['auth.consumer_repository']
+                    true
```
bertramakers: Good catch 👍

app/ApiGuard/ApiGuardServiceProvider.php
```diff
@@ -51,7 +51,7 @@ function (Application $app) {
             }
         );
 
-        $app[self::REPOSITORY] = $app->share(
+        $app[self::CONSUMER_REPOSITORY] = $app->share(
```
bertramakers: Nit: We could just as well use `InMemoryConsumerRepository::class` if we're changing the container alias anyway no? 🤔 But doesn't have to block merging this, we can also clean it up once we move away from Silex and use a better service container with autowiring etc.

PR #48, opened 2019-09-12
Exclude duplicate location from search results by default

bertramakers:
### Added

- Added an `isDuplicate` property on indexed places and events based on the `duplicateOf` property. For events this will currently always be `false` because they never have the `duplicateOf` property in their JSON-LD.
- Added an `isDuplicate=true|false|*` url parameter to include or exclude duplicates. By default this is set to `isDuplicate=false` so duplicates are excluded. However this can be overridden by `disableDefaultFilters=true` or `isDuplicate=*`.

---
Ticket: https://jira.uitdatabank.be/browse/III-3068


src/ElasticSearch/Offer/ElasticSearchOfferQueryBuilder.php
```diff
@@ -417,6 +417,14 @@ public function withOrganizerLabelFilter(LabelName $organizerLabel)
         return $this->withMatchQuery('organizer.labels', $organizerLabel->toNative());
     }
 
+    /**
+     * @inheritDoc
+     */
+    public function withDuplicateFilter(bool $isDuplicate)
```
jdrieghe: 🐜 I prefer the more explicit `withDuplicates()` and `withoutDuplicates()` instead of passing a boolean flag

jdrieghe: I apparently misunderstood the intention of this filter. Can stay as it is for clarity.

PR #50, opened 2019-09-17
Silex end of life

jdrieghe:
Work to do after 07/10:
- [x] Add parallel app folder
- [x] Fix coding standards
- [x] Register /events/ and /places/ routes (use factory to instantiate controller)
- [x] CORS https://github.com/tuupola/cors-middleware
- [x] Adapt ApiKeyRequestAuthenticator to convert PSR-7 to symfony http foundation + add middleware for auth
- [x] Add ResultTransformingPagedCollectionFactoryFactory to switch between embed=true and embed=false output
- [x] Continue to port service providers
- [x] Nice to have: Update udb3-api-guard to use PSR-7 requests

---

Action points discovered on 08/10:

- [x] Error handling
- [x] ~~Find better way to read feature toggles?~~ (Won't do now - Only one feature toggle in this app)
- [x] Remove Silex code and composer dependencies


web/index.php
```diff
@@ -1,35 +1,47 @@
...
+        function () {
+            return Config::load(__DIR__ . '/../config.yml', new Yaml());
+        }
+    );
+    
+    $container->addServiceProvider(LeagueOrganizerServiceProvider::class);
+    
+    $router = new Router();
+    $strategy = (new ApplicationStrategy())->setContainer($container);
+    $router->setStrategy($strategy);
```
jdrieghe: Can we remove this in favour of the `RoutingServiceProvider` you added?

app/Organizer/LeagueOrganizerServiceProvider.php
```diff
@@ -2,22 +2,116 @@
...
+                        new StringLiteral($this->parameter('elasticsearch.organizer.document_type')),
+                        new JsonDocumentTransformingPagedResultSetFactory(
+                            new PassThroughJsonDocumentTransformer(),
+                            new ElasticSearchPagedResultSetFactory(
+                                new NullAggregationTransformer()
+                            )
+                        )
+                    ),
+                    $requestParser,
+                    $this->get('elasticsearch_query_string_factory'),
```
jdrieghe: Can we use `LuceneQueryStringFactory::class` here? Then we don't have to expose and configure the dependency separately later in this file

konrad126: How do you mean? 
To inject the LuceneQueryStringFactory directly?

jdrieghe: Even better: you could inject `QueryStringFactoryInterface::class` here and create a configuration that maps the Lucene implementation to that interface.

web/index.php
```diff
@@ -23,16 +21,12 @@ function () {
             return Config::load(__DIR__ . '/../config.yml', new Yaml());
         }
     );
-    
+
+    $container->addServiceProvider(\CultuurNet\UDB3\SearchService\RoutingServiceProvider::class);
```
jdrieghe: Let's simplify this with a use statement

app/Organizer/LeagueOrganizerServiceProvider.php
konrad126: You men have something like this
`$this->get(QueryStringFactoryInterface::class)` 
?

jdrieghe: It's an option that exposes the service to other services as well.
Alternatively, we can just replace the line with `new LuceneQueryStringFactory()` (I think that is the cleanest for now).

app/Organizer/LeagueOrganizerServiceProvider.php
```diff
@@ -0,0 +1,117 @@
...
+                return new ResultTransformingPagedCollectionFactory(
+                    $this->get('elasticsearch_result_transformer')
+                );
+            }
+        );
+        
+        $this->add(
+            'elasticsearch_result_transformer',
+            function () {
+                return new MinimalRequiredInfoJsonDocumentTransformer();
```
jdrieghe: There is a middleware that overwrites this with `JsonLdEmbeddingJsonDocumentTransformer` when the request contains an `embed` parameter. This might be a bit more complicated to clean up.

app/Organizer/LeagueOrganizerServiceProvider.php
konrad126: The thing is, it is used in other providers like this
`$app['elasticsearch_query_string_factory']`
so I left it like that so others would work

jdrieghe: The other providers aren't booted at this point since both containers aren't compatible with each other. We can't ship this without replacing all of them anyway.

src/Http/ResponseFactory.php
```diff
@@ -1,4 +1,5 @@
-<?php declare(strict_types=1);
+<?php /** @noinspection PhpComposerExtensionStubsInspection */
```
jdrieghe: Why was this added? Can we solve it by either disabling the inspection in your ide or add the necessary composer requirements?

src/Http/Offer/RequestParser/AgeRangeOfferRequestParser.php
```diff
@@ -14,9 +16,9 @@ class AgeRangeOfferRequestParser implements OfferRequestParserInterface
      * @param OfferQueryBuilderInterface $offerQueryBuilder
      * @return OfferQueryBuilderInterface
      */
-    public function parse(Request $request, OfferQueryBuilderInterface $offerQueryBuilder)
+    public function parse(ApiRequestInterface $request, OfferQueryBuilderInterface $offerQueryBuilder)
     {
-        $parameterBagReader = new SymfonyParameterBagAdapter($request->query);
+        $parameterBagReader = new SymfonyParameterBagAdapter(new ParameterBag($request->getQueryParams()));
```
jdrieghe: Let's see if we can clean this up without the dependency on `ParameterBag`

app/Offer/LeagueOfferProvider.php
```diff
@@ -0,0 +1,178 @@
...
+                    new NodeAwareFacetTreeNormalizer(),
+                    new ResultTransformingPagedCollectionFactory(
+                        new MinimalRequiredInfoJsonDocumentTransformer()
+                    )
+                );
+            }
+        );
+    
+        $this->add(
+            'auth.api_key_reader',
```
jdrieghe: Is this service used anywhere else? If not, we can create the key reader in the one location where it's being used

app/ApiKey/ApiKeyReaderSymfonyAdapter.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function __construct(SymfonyApiKeyReaderInterface $apiKeyReader)
+    {
+        $this->apiKeyReader = $apiKeyReader;
+    }
+    
+    public function read(ApiRequestInterface $request)
+    {
+        return $this->apiKeyReader->read(
+            SymfonyRequest::create(
```
jdrieghe: Will we still be needing this or can we refactor the `ApiKeyReader` to read from a PSR-7 compatible request?

konrad126: The `ApiKeyReader` is a package so I guess we would have to update it but then we would need to update other services using it.

I created a wrapper/adapter so we are not dependent on it in our controllers

jdrieghe: Makes sense, we can always rework that part later on. Good call!

web/index.php
```diff
@@ -1,41 +1,17 @@
...
-            return Config::load($configFiles, new Yaml());
-        }
-    );
     
-    $container->addServiceProvider(RoutingServiceProvider::class);
-    $container->addServiceProvider(LeagueOrganizerServiceProvider::class);
-    $container->addServiceProvider(LeagueOfferProvider::class);
-    $container->addServiceProvider(LeagueElasticSearchProvider::class);
+    /** @var Container $container */
+    $container = require __DIR__ . '/../container.php';
```
jdrieghe: Should we do something like `ContainerFactory::create()` instead of requiring a file?

src/Http/OfferSearchController.php
```diff
@@ -91,6 +91,10 @@ class OfferSearchController
      * @var OfferParameterWhiteList
      */
     private $offerParameterWhiteList;
+    /**
```
bertramakers: Nit: There should be a newline before this line @konrad126 . Not sure if the coding standards check will fail on that. Make sure to run those locally for now as long as this PR is a draft to be safe.

src/Http/OfferSearchController.php
```diff
@@ -162,7 +161,7 @@ public function __invoke(ApiRequest $request)
         );
         
         $textLanguages = $this->getLanguagesFromQuery($parameterBag, 'textLanguages');
-
+
```
bertramakers: Same problem as in earlier commits here, make sure to configure PHPStorm so it removes these trailing whitespaces and run the coding standards check locally as long as this PR is a draft.

src/Http/OfferSearchController.php
```diff
@@ -353,7 +356,7 @@ public function __invoke(ApiRequest $request)
             // PagedCollection.
             $jsonArray['facet'][$facetFilter->getKey()] = $this->facetTreeNormalizer->normalize($facetFilter);
         }
-    
+
```
bertramakers: Ditto.

src/Http/Value/Embedded.php
```diff
@@ -0,0 +1,37 @@
+<?php declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search\Http\Value;
+
+class Embedded
```
bertramakers: This value object feels a bit overkill for what is essentially a boolean to be honest. Only the conversion logic is special, but you can just use `$parameterBag->getBooleanFromParameter('embed')` on https://github.com/cultuurnet/udb3-search-service/pull/50/commits/ab89e8dde92bb3e32d406004a03e50c9ceba6a5c#diff-ee98227b8ff9ebc6823659a62b9085abR343

app/CommandServiceProvider.php
```diff
@@ -74,7 +75,9 @@ function () {
                         $this->parameter('elasticsearch.udb3_core_index.reindexation.from'),
                         $this->parameter('elasticsearch.udb3_core_index.reindexation.scroll_ttl'),
                         $this->parameter('elasticsearch.udb3_core_index.reindexation.scroll_size'),
-                        $this->parameter('elasticsearch.udb3_core_index.reindexation.bulk_threshold')
+                        $this->parameter('elasticsearch.udb3_core_index.reindexation.bulk_threshold'),
+                        $this->get(EventBusInterface::class),
+                        $this->get('elasticsearch_indexation_strategy')
```
bertramakers: Won't injecting these services directly will cause them to be instantiated every time you run the console app, even if you run a different command?

app/EventBusProvider.php
```diff
@@ -24,7 +24,25 @@ public function register()
         $this->add(EventBusInterface::class,
             function () {
                 $bus = new SimpleEventBus();
-                // @TODO: add subscribers
+                $bus->beforeFirstPublication(function (EventBusInterface $eventBus) {
+                    $subscribers = [
+                        'organizer_search_projector',
+                        'event_search_projector',
+                        'place_search_projector',
+                    ];
```
bertramakers: Now that we're using a more modern container, it might be more efficient to use tagged definitions instead of having a hardcoded list here? https://container.thephpleague.com/3.x/definitions/ (See "Tagged definitions")

So for example, in `EventServiceProvider`:
```
$this->add(
    'event_search_projector',
    function ()
    {
        return new EventSearchProjector(...);
    }
)->addTag('event_bus_subscribers');
```

And here in `EventBusProvider`:
```
$bus->beforeFirstPublication(function (EventBusInterface $eventBus) {
    // ... 
    foreach ($this->get('event_bus_subscribers') as $subscriber) {
        $eventBus->subscribe($subscriber);
    }
});

File: app/HttpClientProvider.php [hunk 16]
```diff
@@ -0,0 +1,25 @@
+<?php declare(strict_types=1);
+
+namespace CultuurNet\UDB3\SearchService;
+
+use GuzzleHttp\Client;
+
+class HttpClientProvider extends BaseServiceProvider
```
bertramakers: This also needs to override the `protected $provides` property.

app/AmqpProvider.php
```diff
@@ -0,0 +1,99 @@
...
+use CultuurNet\UDB3\Search\Place\PlaceProjectedToJSONLD;
+use PhpAmqpLib\Connection\AMQPStreamConnection;
+use ValueObjects\Number\Natural;
+use ValueObjects\StringLiteral\StringLiteral;
+
+class AmqpProvider extends BaseServiceProvider
+{
+    protected $provides = [
+        'amqp.udb3-core',
+    ];
```
bertramakers: We should override the `provides()` method here instead so we can check if the requested service equals `amqp.$consumerId` like in `register()`.

app/CommandServiceProvider.php
bertramakers: @konrad126 if you have the time, maybe https://symfony.com/doc/3.4/console/lazy_commands.html could help to lazy load (some) commands. Specifically the `ContainerCommandLoader` looks interesting so we can just register the commands as services in the service container and then register them with the CLI app using a simple mapping.

app/EventBusProvider.php
```diff
@@ -0,0 +1,52 @@
...
+                $bus->beforeFirstPublication(function (EventBusInterface $eventBus) {
+                    $subscribers = $this->get('event_bus_subscribers');
+
+                    if (!(is_null($this->parameter('config.event_bus'))) &&
+                        !(is_null($this->parameter('config.event_bus.subscribers')))) {
+
+                        $subscriberIds = $this->parameter('config.event_bus.subscribers');
+                        $subscribers = [];
+                        foreach ($subscriberIds as $subscriberServiceId) {
+                            $subscribers[] = $subscriberServiceId;
```
bertramakers:
```suggestion
                            $subscribers[] = $this->get($subscriberServiceId);
```
The ids are just strings from the config file so we should fetch the actual services from the service container before subscribing them.

bertramakers:
Everything seems to work as before on acceptance, so this can be merged!

PR #51, opened 2019-10-08
III-3080 Index location.mainId on event documents

bertramakers:
This only contains the id of the location that is actually linked to the event, and not the duplicate/master ids.

### DIFF
--- a/tests/ElasticSearch/Event/data/indexed-without-start-date.json
+++ b/tests/ElasticSearch/Event/data/indexed-without-start-date.json
@@ -29,6 +29,7 @@
         "@id": "http:\/\/udb-silex.dev\/place\/179c89c5-dba4-417b-ae96-62e7a12c2405",
         "@type": "Place",
         "id": "179c89c5-dba4-417b-ae96-62e7a12c2405",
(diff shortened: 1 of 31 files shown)


tests/ElasticSearch/Event/data/indexed-without-start-date.json
```diff
@@ -29,6 +29,7 @@
         "@id": "http:\/\/udb-silex.dev\/place\/179c89c5-dba4-417b-ae96-62e7a12c2405",
         "@type": "Place",
         "id": "179c89c5-dba4-417b-ae96-62e7a12c2405",
+        "mainId": "179c89c5-dba4-417b-ae96-62e7a12c2405",
```
brampauwelyn: What's the difference between `id` & `mainId`?

bertramakers: Since the "mark as duplicate" project the `location.id` property can contain multiple ids, those of the master location and the duplicate location(s). The `location.mainId` property will only contain the id of the location that is actually linked to the event, and not the "related" locations. That makes it easier to find only the events that are linked to a specific location and not its canonical / duplicate id(s).

PR #54, opened 2019-11-07
Make sure availableTo is never lower than availableFrom

bertramakers:
### Fixed

- Fixed some (test) events and places not being indexed because their `availableFrom` was higher than `availableTo`. There's no real use case for this, but hinders testing from external developers. To fix it we set `availableTo` to the `availableFrom` date if it's lower.

---

Ticket: https://jira.uitdatabank.be/browse/III-2877


bertramakers:
All changes should be compatible with the silex EOL branch, so after merging we should be able to cherry-pick the commits to that branch.

bertramakers:
I've also moved some code around to better structure it as part of an ongoing refactor of this part of the codebase. For easier reviewing, review the commits individually.

tests/ElasticSearch/Event/EventJsonDocumentTransformerTest.php
```diff
@@ -462,6 +462,30 @@ public function it_uses_endDate_if_availableTo_is_missing()
         $this->assertEquals($expectedLogs, $actualLogs);
     }
 
+    /**
+     * @test
+     */
+    public function it_uses_makes_sure_availableFrom_is_never_higher_than_availableTo(): void
```
brampauwelyn: maybe a bit too long? 
why it_uses_makes_sure... and not just it_makes_sure... ?

bertramakers: Yep good point, I copy-pasted the test before it and didn't change the name correctly. Should be fixed now!

PR #62, opened 2020-09-22
Technical documentation

bertramakers:
### Added

- Added documentation on how to index new fields and add URL parameters for them
- Added PR template

---

Ticket: https://jira.uitdatabank.be/browse/III-3473


readme.md
```diff
@@ -1 +1,147 @@
...
+```bash
+./bin/app.php elasticsearch:migrate
+```
+
+An actual migration will only occur if the script detects a new version number in the document mapping. At that point, it will create a new index and configure it with the latest mapping, and then re-index all documents from the old index by looping over them and fetching the latest JSON-LD for each one from UDB3 and indexing that in the new index.
+
+This means that this command is idempotent. You can run it as much as you want without doing any checks beforehand, e.g. after every deploy or git pull.
+
+
+To keep the search index live while re-indexing in production (and other environments), we work with two aliases:
```
LucWollants, line 30: The documentation reads nicely! I would just add another header here "How does it work"

LucWollants (approved):
Really nice documentation 👌

jdrieghe (approved):

readme.md
bertramakers: Thanks for the tip, added it!

PR #65, opened 2020-09-23
Phpstan level 2

bertramakers:
### Changed

- Changed phpstan from level 1 to 2

### Fixed

- Fixed phpstan level 2 errors


LucWollants (approved):
What PHPStan level do you aim for? 😀

bertramakers:
> Which PHPStan level do you aim for? 😀

We didn't really discuss that yet, but I'd say 8 😄 
But so far we always stopped at a level where the errors became too much / too hard to fix right now.

PR #67, opened 2020-09-23
III-3473 Production id filter

bertramakers:
### Added
 
- Added production id filter
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3473

---

Todo:

- [X] Add field mapping for `production.id`
- [X] Increase `UDB3_CORE` version number
- [X] Add logic to copy production id from incoming json-ld to elasticsearch json document
- [x] Add `withProductionId` filter to `OfferQueryBuilderInterface`
- [x] Implement `withProductionId` in `ElasticSearchOfferQueryBuilder`
- [x] Add request parser for `productionId` URL parameter
- [x] Inject new request parser in `OfferSearchController`

PR #68, opened 2020-09-24
Fix ElasticSearch migrate issues

bertramakers:
### Fixed
 
- Fixed some documents not being indexed in bulk but instead file per file when migrating to a new index, because the indexation strategy was not a singleton.
- Fixed the event bus not being a singleton and being instantiated multiple times.
- Fixed events and places not being re-indexed when migrating to a new index. The necessary messages were dispatched on the internal event bus, but the projectors for indexing events and places were not subscribed correctly.

---

Noticed this while working on #67 and testing the new field mapping

### DIFF
--- a/app/Event/EventServiceProvider.php
+++ b/app/Event/EventServiceProvider.php
@@ -15,6 +15,7 @@ use ValueObjects\StringLiteral\StringLiteral;
 class EventServiceProvider extends BaseServiceProvider
 {
     protected $provides = [
+        Client::class,
         'event_controller',
         'event_search_projector',
         'event_bus_subscribers',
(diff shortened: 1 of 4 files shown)


app/Event/EventServiceProvider.php
```diff
@@ -15,6 +15,7 @@
 class EventServiceProvider extends BaseServiceProvider
 {
     protected $provides = [
+        Client::class,
```
bertramakers, line 18: I'm not sure why this fixes the `event_search_projector` not being subscribed when running `elasticsearch:migrate`. But it was the only difference with `OrganizerServiceProvider`, where `organizer_search_projector` was defined and successfully subscribed. Extra weird is that all projectors were already being subscribed correctly when running `consume-udb3-core` instead...

bertramakers: It seems like some weird bug/logic in the League container when working with service tags. The tag is defined as provided as well, but it seems like the provider was not bootstrapped when running `elasticsearch:migrate` specifically.

jdrieghe (approved):

PR #70, opened 2020-09-29
III-2201 Remove i18n fallback code 🧹🌐

bertramakers:
### Removed
 
- Removed i18n fallback logic for missing properties and outdated property formats, added in the past to handle changes made to the read models while not being able to do a full replay to make all the data consistent
 
---

Ticket: https://jira.uitdatabank.be/browse/III-2201


bertramakers:
The last commit might be slightly messy because I forgot to commit one or two minor fixes to the tests related to previous fallbacks that were removed. 99% of it is actually related to the last fallback that was removed though.

src/ElasticSearch/JsonDocument/CopyJson/Components/CopyJsonLanguages.php
```diff
@@ -2,25 +2,22 @@
 
 namespace CultuurNet\UDB3\Search\ElasticSearch\JsonDocument\CopyJson\Components;
 
+use CultuurNet\UDB3\Search\ElasticSearch\JsonDocument\CopyJson\Logging\CopyJsonLoggerInterface;
 use CultuurNet\UDB3\Search\Language\Language;
 use CultuurNet\UDB3\Search\Language\JsonDocumentLanguageAnalyzer;
-use CultuurNet\UDB3\Search\ReadModel\JsonDocument;
 use CultuurNet\UDB3\Search\ElasticSearch\JsonDocument\CopyJson\CopyJsonInterface;
 use Rhumsaa\Uuid\Uuid;
```
LucWollants, line 9: It seems there are a few more unused imports.

src/ElasticSearch/JsonDocument/CopyJson/Components/CopyJsonName.php
```diff
@@ -28,13 +28,6 @@ public function copy(\stdClass $from, \stdClass $to)
     {
         $mainLanguage = isset($from->mainLanguage) ? $from->mainLanguage : 'nl';
```
LucWollants, line 29: I think the `$mainLanguage` variable is no longer used.

src/ElasticSearch/Offer/AbstractOfferJsonDocumentTransformer.php
```diff
@@ -28,27 +28,19 @@ abstract class AbstractOfferJsonDocumentTransformer implements JsonDocumentTrans
...
-    /**
-     * @var JsonDocumentLanguageAnalyzer
-     */
-    protected $languageAnalyzer;
-
     /**
      * @param IdUrlParserInterface $idUrlParser
      * @param OfferRegionServiceInterface $offerRegionService
      * @param LoggerInterface $logger
-     * @param JsonDocumentLanguageAnalyzer $languageAnalyzer
```
LucWollants, line 40: This can now also be removed from the imports.

src/ElasticSearch/JsonDocument/CopyJson/Components/CopyJsonLanguages.php
bertramakers: Yes true, but I think they were already unused before my changes. In any case, I already cleaned them up in #71 and that also refactors these CopyJson classes quite a bit (but it's still wip). Maybe best to not clean them up here as well then, because it would create a lot of merge conflicts?

src/ElasticSearch/JsonDocument/CopyJson/Components/CopyJsonName.php
bertramakers: True, good point. But as mentioned in my previous comment I'm refactoring these classes in #71 . Okay if I remove it there?

src/ElasticSearch/Offer/AbstractOfferJsonDocumentTransformer.php
bertramakers: Yep. But again it will create a lot of merge conflicts with #71 (and it's already cleaned up there)

LucWollants (approved):
Nice cleanup! I like PHPStan 😉

PR #71, opened 2020-09-29
III-3474 include regions in response json

bertramakers:
### Added
 
- Added `regions`, which is calculated by SAPI3, to embedded json-ld results
 
### Changed
 
- Refactored multiple approaches of how to transform JSON to a single approach
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3474

This PR includes a lot of refactoring and cleanup to consolidate all the various ways we had to copy/transform JSON. The diff has gotten quite big so it's best to review commit by commit.

Essentially everything up to and including 360fbe7 is to make the extra `regions` property work. Everything after that is refactoring and clean up of existing code.

Here's why I did this extra work:
- The `ResultTransformerFactory` used the `JsonDocumentTransformerInterface` to transform the Elasticsearch result to the response JSON. However, this interface only supports passing in a "from" and then getting a new document. But to combine the regions calculated by Elasticsearch together with the originally embedded JSON-LD also on the Elasticsearch result without having a class that did two things at the same time, we needed a way to pass in a "draft" document where the transformers can make their changes on in an immutable way. This is what the new `JsonTransfomer` interface does. 
- Other `JsonDocumentTransformerInterface` implementations already used a similar concept of `CopyJsonInterface` underneath, but this used mutable `stdClass` arguments instead of an immutable approach. The lines between `JsonDocumentTransformerInterface` and `CopyJsonInterface` were also very fuzzy.
- Instead of introducing yet another standard, I thought it would be a good opportunity to clean up the following existing three ways we had to convert JSON to a new single approach:
  - `JsonDocumentTransformerInterface`
  - `CopyJsonInterface`
  - Re-usable protected methods on `AbstractOfferJsonDocumentTransformer`

As an extra bonus, I was able to remove several layers of abstractions here and there.


src/ElasticSearch/JsonDocument/JsonLdEmbeddingJsonTransformer.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace CultuurNet\UDB3\Search\ElasticSearch\JsonDocument;
+
+use CultuurNet\UDB3\Search\JsonDocument\JsonTransformer;
+
+class JsonLdEmbeddingJsonTransformer implements JsonTransformer
+{
+    public function transform(array $original, array $draft = []): array
+    {
+        $originalJsonLd = json_decode($original['originalEncodedJsonLd'] ?? '{}', true);
```
jdrieghe, line 11: Should we add the `JSON_THROW_ON_ERROR` option here in case decoding fails?

bertramakers: Yes, very good idea! 👍

src/ElasticSearch/JsonDocument/MinimalRequiredInfoJsonTransformer.php
```diff
@@ -0,0 +1,20 @@
...
+ * Converts Event, Place and Organizer results to minimal documents that only
+ * contain @id and @type.
+ * Should be used when returning search results.
+ */
+class MinimalRequiredInfoJsonTransformer implements JsonTransformer
+{
+    public function transform(array $original, array $draft = []): array
+    {
+        $draft['@id'] = $original['@id'] ?? '';
+        $draft['@type'] = $original['@type'] ?? '';
```
jdrieghe, line 17: `$draft` was marked as immutable here: https://github.com/cultuurnet/udb3-search-service/pull/71/commits/5177237a6c49686300d7af8df95132ab34592ab0#diff-1c3f56cfec207a5e7578dc82fdde484dR11

Now we are mutating it. Should this be something like this instead?
```
return [
   '@id' => $original['@id'] ?? '',
   '@type' => $original['@type'] ?? '',
]
```

tests/ElasticSearch/JsonDocument/MinimalRequiredInfoJsonTransformerTest.php
```diff
@@ -0,0 +1,43 @@
...
+            'foo' => 'bar',
+            'lorem' => 'ipsum',
+        ];
+
+        $expected = [
+            '@id' => 'http://foo.io/events/b1a44077-722b-4a72-9621-50cb8dbb46db',
+            '@type' => 'Event',
+        ];
+
+        $actual = $this->transformer->transform($original);
```
jdrieghe, line 37: You're missing a test where you pass a `$draft` with actual data, which would fail the expectations I think.

src/ElasticSearch/JsonDocument/MinimalRequiredInfoJsonTransformer.php
bertramakers: The idea is to merge the existing data on `$draft` (if any) with extra properties. We could do that with `array_merge`, but this also works because arrays are not passed by reference unless you use `&` in the argument name (like `&$draft`).

tests/ElasticSearch/JsonDocument/MinimalRequiredInfoJsonTransformerTest.php
bertramakers: I'll add the test, it should prove/disprove my reply to your previous comment.

tests/ElasticSearch/JsonDocument/RegionEmbeddingJsonTransformerTest.php
```diff
@@ -0,0 +1,68 @@
...
+use PHPUnit\Framework\TestCase;
+
+final class RegionEmbeddingJsonTransformerTest extends TestCase
+{
+    /**
+     * @var RegionEmbeddingJsonTransformer
+     */
+    private $transformer;
+
+    public function setUp(): void
```
jdrieghe, line 16: PHPUnit's setUp method is `protected`. No need to make it `public` here.

src/ElasticSearch/JsonDocument/Properties/IdentifierTransformer.php
```diff
@@ -40,30 +40,27 @@ public function __construct(
...
-     * @param \stdClass $from
-     * @param \stdClass $to
-     */
-    public function copy(\stdClass $from, \stdClass $to)
+    public function transform(array $from, array $draft = []): array
     {
-        if (isset($from->{"@id"})) {
-            $to->{"@id"} = $from->{"@id"};
+        if (isset($from["@id"])) {
+            $draft["@id"] = $from["@id"];
```
jdrieghe, line 46: Nitpicking, but we should use single quotes for these. Something we can fix with automatic code style checks?

bertramakers: I agree, it's probably left over from the previous code (looking at the diff). We could add a code style check for that yes, but we should probably do that in another PR later because we should also update CodeSniffer in various projects and get the style rules more consistent.

bertramakers: @LucWollants added https://jira.uitdatabank.be/browse/III-3529 for that. I've also created a filter for technical debt tickets that have not been picked up yet. That way we can keep track of them and then plan them gradually instead of trying to improve everything at once: https://jira.uitdatabank.be/browse/III-3529?filter=12988

src/ElasticSearch/JsonDocument/PlaceJsonDocumentTransformer.php
```diff
@@ -36,12 +36,18 @@ public function __construct(
...
-        $newBody = new \stdClass();
 
         $this->logger->debug("Transforming place {$id} for indexation.");
 
-        $this->placeTransformer->copy($body, $newBody);
+        // @todo refactor copy methods to transformer classes and remove workaround to make newBody an stdClass
+        $from = json_decode($jsonDocument->getRawBody(), true);
+        $to = [];
+        $body = $jsonDocument->getBody();
+        $newBody = json_decode(
```
jdrieghe: Isn't it possible to cast the array to object instead of going through `decode(encode())`?

bertramakers: I'll double-check but I think it only casts the top level, not nested arrays.

bertramakers: Oh but this was just a temporary workaround while refactoring to keep the tests working, which is removed later

jdrieghe (approved):

tests/ElasticSearch/JsonDocument/MinimalRequiredInfoJsonTransformerTest.php
bertramakers: Added the test in https://github.com/cultuurnet/udb3-search-service/pull/71/commits/adf48351f9af682492d010809d6c6bfb647672ff

src/ElasticSearch/JsonDocument/MinimalRequiredInfoJsonTransformer.php
bertramakers: The test added in https://github.com/cultuurnet/udb3-search-service/pull/71/commits/adf48351f9af682492d010809d6c6bfb647672ff should show that the value that's passed as `$draft` does not get mutated outside of the method.

src/ElasticSearch/JsonDocument/JsonLdEmbeddingJsonTransformer.php
bertramakers: I tried adding it but it's only available in PHP 7.3+, and we still need compatibility with 7.1 for obvious reasons sadly.
I could check the return value and make it throw if it's `null`, but then again it would "break" the whole page of results if only one results JSON-LD is invalid. It would also be very unlikely to happen since if it's not valid, it couldn't have been decoded to create the Elasticsearch JSON document from the JSON-LD in the first place. So I don't think it's worth adding that check right now?

jdrieghe: Nah, it's fine like this then.

jdrieghe (approved):

jdrieghe:
🚢  it!

PR #72, opened 2020-09-30
III-3386 Allow throwable for API report creation

LucWollants:
### Fixed
 
- Allow throwable type as parameter for api problem report creation
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3386

### DIFF
--- a/app/Error/ApiExceptionHandler.php
+++ b/app/Error/ApiExceptionHandler.php
@@ -49,7 +49,7 @@ class ApiExceptionHandler extends Handler
         return Handler::QUIT;
     }
 
-    private function createNewApiProblem(\Exception $e)
+    private function createNewApiProblem(\Throwable $e)
     {
         $problem = new ApiProblem($e->getMessage());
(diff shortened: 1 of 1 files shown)


jdrieghe (approved):

app/Error/ApiExceptionHandler.php
```diff
@@ -49,7 +49,7 @@ public function handle()
         return Handler::QUIT;
     }
 
-    private function createNewApiProblem(\Exception $e)
+    private function createNewApiProblem(\Throwable $e)
```
bertramakers, line 52: Won't we be leaking potential security issues like this, by converting _every_ throwable to an API problem and including the message?
The API problem response (or at least the messages) was only intended to be used for user errors.
So I would either not create a response for it and let the Throwable cascade and result in a white page of death, or only include the Throwable message here if it's not an `Error` class and otherwise include a generic "Something went wrong" message.

LucWollants: That is indeed an issue. I created a PR to solve the extra `Error`s we could expose: https://github.com/cultuurnet/udb3-search-service/pull/74

PR #76, opened 2020-10-13
III-3476 Add option to group events by production id

bertramakers:
### Added
 
- Added option to group events (or offers if searching on both events + places) by production id. Events without production id, and places, are never grouped together. (Technically the "grouping" excludes all events with a production id that already appeared in the results. But those events are also linked in the first event under the `production.otherEvents` property)
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3476

---

Note: The package we use to build the Elasticsearch query does not support field collapsing. It was easy to add it to the query array, but it required some minor refactoring.

I also had to add a workaround to get the correct `total` because field collapsing does not change the `total` returned by Elasticsearch. So instead I also added an aggregation on a distinct value for the field we collapse on (`productionCollapseValue`) which will return the correct `total`. So when collapsing, we use that total instead. Otherwise we use the normal `total`.

See: https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-request-collapse.html

> The total number of hits in the response indicates the number of matching documents without collapsing. The total number of distinct group is unknown.


src/ElasticSearch/ElasticSearchPagedResultSetFactory.php
```diff
@@ -56,14 +56,15 @@ function (array $result) {
         );
 
         $aggregations = isset($response['aggregations']) ? $response['aggregations'] : [];
-        array_walk(
-            $aggregations,
-            function (array &$aggregationData, $aggregationName) {
-                $aggregationData = Aggregation::fromElasticSearchResponseAggregationData(
+        $aggregations = array_map(
+            function (array $aggregationData, $aggregationName) {
```
jdrieghe:
```suggestion
            function (array $aggregationData, $aggregationName): Aggregation {
```

jdrieghe: Can we type $aggregationName as well?

src/ElasticSearch/ElasticSearchPagedResultSetFactory.php
```diff
@@ -56,14 +57,28 @@ function (array $result) {
...
-                );
-            }
+
+        if (isset($aggregations['total'])) {
+            $total = new Natural($aggregations['total']['value']);
+        }
+
+        $bucketAggregations = array_filter(
+            array_map(
+                function (array $aggregationData, $aggregationName) {
```
jdrieghe, line 67:
```suggestion
                function (array $aggregationData, $aggregationName): Aggregation {
```

jdrieghe (approved):
Thanks for the clear commits and extra inline comments, they really clarified things while reviewing 👍

src/ElasticSearch/ElasticSearchPagedResultSetFactory.php
bertramakers: Yes, good suggestions!

bertramakers:
@jdrieghe I've added another commit with your suggestions for the typing in the `array_map` function. Thanks for the positive feedback! I paid extra attention to it with your earlier feedback in mind ;)

jdrieghe (approved):

src/ElasticSearch/ElasticSearchPagedResultSetFactory.php
```diff
@@ -64,7 +64,7 @@ function (array $result) {
 
         $bucketAggregations = array_filter(
             array_map(
-                function (array $aggregationData, string $aggregationName): ?Aggregation {
+                function (array $aggregationData, string $aggregationName) : ?Aggregation {
```
jdrieghe, line 67: This is a really weird rule 🤔

jdrieghe (approved):

src/ElasticSearch/ElasticSearchPagedResultSetFactory.php
bertramakers: Yes it is

bertramakers: Anyway, it should improve when we pick up the ticket to update our coding standards rulesets and tooling.

PR #80, opened 2020-10-29
III-3545 Add sorting on popularity

LucWollants:
### Added
 
- Add sorting on popularity
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3545


src/ElasticSearch/JsonDocument/Properties/MetadataTransformer.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    public function transform(array $from, array $draft = []): array
+    {
+        $metadata = $this->getMetadata($from);
+
+        if (empty($metadata)) {
+            return $draft;
+        }
+
+        $draft['metadata'] = $metadata;
```
bertramakers, line 19: Nit: You could remove the whole `$this->getMetadata()` method and `empty()` check above by chaging this to:
```
$draft['metadata'] = [
  'popularity' => $from['metadata']['popularity'] ?? 0,
];
```

LucWollants: This will also add a `popularity` score of `0` to every indexed event and place, but that is what we want for sorting.
I will fix the tests to take this new property into account.

bertramakers: Yeah. I also think for sorting on integer fields, `null` (or a missing field) and `0` is the same for Elasticsearch. So we might as well set it to `0` anyway. Plus, normally every document should have a `metadata.popularity` on udb3 now (because we also default it to `0` there).

src/JsonDocument/QueryJsonDocument.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search\JsonDocument;
+
+class QueryJsonDocument
```
bertramakers, line 7: Nit: `Query` might become confusing with the `QueryBuilder` interfaces and classes (and just the concept of querying in search). So maybe `FetchJsonDocument` might be clearer?

bertramakers: Oh wait I was under the assumption that this would do the actual fetching. But instead it's just a value object so I understand why it has the name `Query`.

Still, might be clearer and a bit more cohesive if this class would be something like `JsonDocumentFetcher` (name is just an example) which gets a Guzzle client injected and does the actual fetching (with or without metadata parameter)? And then this could be injected in `TransformingJsonDocumentIndexService` instead of the Guzzle client?

bertramakers: Aside from this minor suggestion, everything looks great 👍

LucWollants: Proposal [pull request #81](https://github.com/cultuurnet/udb3-search-service/pull/81) for the new `JsonDocumentFetcher` class.

bertramakers (approved):
Approving here as well because the only suggestion I had is solved by #81

bertramakers:
Very nice work @LucWollants 👌 💯

jdrieghe (approved):

PR #86, opened 2021-02-08
III-3782 Internalize hydra

LucWollants:
### Changed
 
- Merge `cultuurnet/hydra` dependency inside `udb3-search-service`

---

Ticket: https://jira.uitdatabank.be/browse/III-3782


bertramakers (approved):
Looks great! 👍 

I noticed there was a `PagedCollectionTest` that wasn't copied over. https://github.com/cultuurnet/Hydra/blob/master/test/PagedCollectionTest.php
Maybe that was intentional, or not. In any case I don't think it's necessary because you also removed a lot of unnecessary functionality and what's left over is covered by the other unit tests in this repo I think. But just wanted to mention it in case you forgot and think differently.

Thanks for the extra cleanup like adding the `ext-json` and removing the `PageUrlGenerator` etc! ✨

LucWollants:
> Maybe that was intentional, or not.

Indeed it was intentional not to copy `PagedCollectionTest`. But it was unintentional not to mention it inside the PR description.

PR #87, opened 2021-02-08
III-3568 Refactor filtering on status

bertramakers:
### Changed
 
- Changed how filtering on status happens when used in combination with dateFrom/dateTo. This approach makes it easier to combine with other subEvent properties like the new timeRange which will be added in https://jira.uitdatabank.be/browse/III-3431
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3568

For more info, see https://www.monterail.com/blog/how-to-index-objects-elasticsearch


jdrieghe (approved):
Wow, very nice work here @bertramakers

PR #88, opened 2021-02-10
III-3706 Replace phing with composer scripts

jdrieghe:
### Removed
- dependency on `phing/phing` in favour of separate composer scripts in the CI pipeline
 
---
Ticket: https://jira.uitdatabank.be/browse/III-3706


bertramakers (approved):
Looks good! 👍  For the remaining repositories I wouldn't try to optimize the coding standard config too much. Renaming the ruleset file so it gets picked up automatically is necessary of course, but changing the rules now to make them consistent matters very little because we'll switch to php-cs-fixer with a different ruleset soon anyway. (It's on the board if I'm not mistaken)

jdrieghe:
@bertramakers just using the same ruleset everywhere (even when enforced through phpcs) will make it easier to migrate to a unified php-cs-fixer ruleset later, no?

bertramakers:
Depends on whether we use the same ruleset in php-cs-fixer I think?
And I assume that php-cs-fixer would make it easier to switch to a new ruleset, if it can fix errors automatically?

bertramakers:
But if it's not much work and you think it would help definitely go for it! I just wanted to make sure you're aware we're switching soon anyway 😄

jdrieghe:
@bertramakers `phpcbf` can also fix errors indicated by `phpcs` automatically, which is what it did here.

LucWollants:
I am also in favor of upgrading phpcs to version 3.5, because this will fix the PHPStorm hangups I sometimes experience.

LucWollants (approved):

PR #92, opened 2021-02-11
III-3436 Add filter on local time range

bertramakers:
### Added
 
- Added a way to filter on the local time that an event starts/ends or a place opens/closes. Indexing it as a date field with the `H:i` format caused too much problems, for example using it in the `q` parameter was difficult because Elasticsearch still assigns a date to it and you'd have to do use the complete date notation in the `q` parameter. See the ticket & comments for more info on these issues.
Instead we index it as an integer between `0` (`00:00`) and `2359` (`23:59`). So the usage with the URL parameter is `localTimeFrom=0800&localTimeTo=1600` for example (or without the leading 0 in 0800). 
In the `q` parameter the equivalent is `q=localTimeRange:[800 TO 1600]`. 
Before indexing the time info, the time is converted to the timezone of the country the event takes place in (hence "local").
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3436

More documentation can be found in this comment on Jira: https://jira.uitdatabank.be/browse/III-3436?focusedCommentId=107541&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-107541 (Will be added to documentatie.uitdatabank.be by Corneel after it's deployed)

The +7K lines is caused by the test data for permanent/period events and places with opening hours, because they get subEvents for every instance of the opening hours from 6 months in the past to 2 years in the future and those subEvents now also get a `localTimeRange` property. Probably best to review this commit-per-commit so you can easily skip the test data changes.


src/ElasticSearch/JsonDocument/Properties/CalendarTransformer.php
```diff
@@ -298,6 +314,64 @@ private function convertSubEventToDateRange(array $subEvent): stdClass
         );
     }
 
+    private function convertSubEventsToLocalTimeRanges(array $subEvents): array
```
jdrieghe: I prefer to add additional docblock type hints for `array` to clarify the types held by the `array`

src/ElasticSearch/JsonDocument/Properties/CalendarTransformer.php
```diff
@@ -14,6 +14,17 @@
...
 final class CalendarTransformer implements JsonTransformer
 {
+    /**
+     * List of countries that UDB3 supports and their timezones so we can index localTimeRange based on the start and
+     * end times of an event converted to the timezone in which it takes place.
+     * @see https://github.com/eggert/tz/blob/master/zone1970.tab
+     */
+    private const TIMEZONES = [
+        'BE' => 'Europe/Brussels',
+        'NL' => 'Europe/Amsterdam',
```
jdrieghe, line 24: Are we planning on being in a timezone that's *actually* different?
If the answer is no, we may have added more abstraction and "flexibility" here than is needed to solve the problem.

src/ElasticSearch/Offer/ElasticSearchOfferQueryBuilder.php
```diff
@@ -158,11 +159,12 @@ function (Status $status) {
         );
     }
 
-    public function withStatusAwareDateRangeFilter(
-        \DateTimeImmutable $from = null,
-        \DateTimeImmutable $to = null,
-        Status ...$statuses
-    ) {
+    public function withSubEventFilter(SubEventQueryParameters $subEventQueryParameters)
```
jdrieghe, line 168: Can we add a return type here?

src/ElasticSearch/JsonDocument/Properties/CalendarTransformer.php
bertramakers: Yes but I will clarify on Slack because I don't know if the specific information is allowed to be public.

jdrieghe (approved):

PR #98, opened 2021-03-02
III-3841 Mark classes as final (Part 2)

LucWollants:
### Changed
- Made `AbstractReindexCommand` abstract
- Made `JSONDeserializer` abstract
- Marked `BulkIndexationStrategy` as final
- Marked `JsonDocumentFetcher` as final
- Marked `FacetName` as final

### Removed
- Remove test for unknown `FacetName` passed to `ElasticSearchOfferQueryBuilder::withFacet`

---

Ticket: https://jira.uitdatabank.be/browse/III-3841


src/Deserializer/JSONDeserializer.php
```diff
@@ -6,7 +6,7 @@
 
 use ValueObjects\StringLiteral\StringLiteral;
 
-class JSONDeserializer implements DeserializerInterface
+abstract class JSONDeserializer implements DeserializerInterface
```
bertramakers, line 9: There's only one class extending this, `DomainMessageJSONDeserializer`. Maybe we should just integrate it in there and remove this abstract parent class?

Some extra context:
It was split up in the past because this abstract class was in the `cultuurnet/deserializer` package, and the `DomainMessageJSONDeserializer` implementation was part of `cultuurnet/udb3` if I'm not mistaken (but moved a while back already to remove the `cultuurnet/udb3` dependency in sapi3).

src/JsonDocument/JsonDocumentFetcher.php
```diff
@@ -8,7 +8,7 @@
 use GuzzleHttp\ClientInterface;
 use Psr\Log\LoggerInterface;
 
-class JsonDocumentFetcher
+final class JsonDocumentFetcher implements JsonDocumentFetcherInterface
```
bertramakers, line 11: Maybe we can rename this to something like `GuzzleJsonDocumentFetcher` and `JsonDocumentFetcher` (without the `Interface` suffix)?

bertramakers (commented):
Nice refactoring! I just have some minor suggestions

tests/ElasticSearch/IndexationStrategy/MutableIndexationStrategyTest.php
```diff
@@ -36,7 +36,7 @@ protected function setUp()
         $this->mockStrategy1 = $this->createMock(IndexationStrategyInterface::class);
         $this->mockStrategy2 = $this->createMock(IndexationStrategyInterface::class);
 
-        $this->mockBulkStrategy = $this->createMock(BulkIndexationStrategy::class);
+        $this->mockBulkStrategy = $this->createMock(BulkIndexationStrategyInterface::class);
```
bertramakers, line 39: If the interface was added just to make this mocking work, would it also work if we didn't mock the `BulkIndexationStrategy` but instead mocked the dependencies inside its constructor?

Because I don't really see a second kind of `BulkIndexationStrategyInterface` popping up in the future.

bertramakers: Alternatively, we could move the `flush()` to the parent interface and call it something like `finish`? Because it's only called when the last `indexDocument()` call has been done, to make sure the last ones also get indexed. Everything in-between gets "auto-flushed" inside `BulkIndexationStrategy` itself after every x documents.

So I don't see a lot of harm in adding a `finish()` method to `IndexationStrategyInterface` and just not make it do anything in the other implementations.

The fact that we're calling this `BulkIndexationStrategyInterface` and `BulkIndexationStrategy` just makes me think that this interface should not be necessary. 🤔

src/Deserializer/JSONDeserializer.php
LucWollants: If `DomainMessageJSONDeserializer` is the only one using this `JSONDeserializer` it indeed makes sense to remove `JSONDeserializer`

src/JsonDocument/JsonDocumentFetcher.php
LucWollants: I also have the habit now of not adding the `Interface` suffix, but I had the impression it was common in this project to use the suffix?

tests/ElasticSearch/IndexationStrategy/MutableIndexationStrategyTest.php
LucWollants: > If the interface was added just to make this mocking work, would it also work if we didn't mock the BulkIndexationStrategy but instead mocked the dependencies inside its constructor?

Yes, this was a fast workaround for the mocking

> So I don't see a lot of harm in adding a finish() method to IndexationStrategyInterface and just not make it do anything in the other implementations.

I think this solution has the "best of both worlds" so I would go for this approach?

src/JsonDocument/JsonDocumentFetcher.php
bertramakers: Yes but we should probably move away from that. Preferably in small incremental changes over time of course.

LucWollants: I already renamed a few low "hanging fruits". But the rest will have to wait for another PR.

bertramakers: I think we can also drop it in `IndexationStrategyInterface`?

LucWollants: Done

bertramakers (approved):
Nice!

PR #99, opened 2021-03-04
III-3829

jdrieghe:
### Added
- Calendar summaries can now be embedded by passing `?embed=true&embedCalendarSummaries[]='xs-text'`. Multiple can be combined and these are the allowed values:
```
xs-text
sm-text
md-text
lg-text
xs-html
sm-html
md-html
lg-html
```

---

Ticket: https://jira.uitdatabank.be/browse/III-3829

### DIFF
--- a/src/ElasticSearch/JsonDocument/CalendarSummaryEmbeddingJsonTransformer.php
+++ b/src/ElasticSearch/JsonDocument/CalendarSummaryEmbeddingJsonTransformer.php
@@ -4,8 +4,13 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Search\ElasticSearch\JsonDocument;
 
+use CultuurNet\CalendarSummaryV3\CalendarFormatterInterface;
+use CultuurNet\CalendarSummaryV3\CalendarHTMLFormatter;
+use CultuurNet\CalendarSummaryV3\CalendarPlainTextFormatter;
+use CultuurNet\CalendarSummaryV3\Offer\Offer;
 use CultuurNet\UDB3\Search\JsonDocument\JsonTransformer;
 use CultuurNet\UDB3\Search\Offer\CalendarSummaryFormat;
+use InvalidArgumentException;
 
 class CalendarSummaryEmbeddingJsonTransformer implements JsonTransformer
 {
@@ -30,9 +35,17 @@ class CalendarSummaryEmbeddingJsonTransformer implements JsonTransformer
 
     private function embedCalendarSummary(array $original, CalendarSummaryFormat $calendarSummaryFormat): array
     {
+        $calendarFormatter = $this->getCalendarFormatterByType($calendarSummaryFormat->getType());
+        $offer = Offer::fromJsonLd(json_encode($original));
+
         $calendarSummary = [
             $calendarSummaryFormat->getType() => [
-                $calendarSummaryFormat->getFormat() => '',
+                $calendarSummaryFormat->getFormat() => trim(
+                        $calendarFormatter->format(
+                        $offer,
+                        $calendarSummaryFormat->getFormat()
+                    )
+                ),
             ]
         ];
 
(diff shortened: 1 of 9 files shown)


src/ElasticSearch/JsonDocument/CalendarSummaryEmbeddingJsonTransformer.php
```diff
@@ -0,0 +1,71 @@
...
+
+    public function __construct(CalendarSummaryFormat ...$calendarSummaryFormats)
+    {
+        $this->calendarSummaryFormats = $calendarSummaryFormats;
+    }
+
+    public function transform(array $original, array $draft = []): array
+    {
+        foreach ($this->calendarSummaryFormats as $calendarSummaryFormat) {
+            $original = $this->embedCalendarSummary($original, $calendarSummaryFormat);
```
LucWollants, line 30: Should this be copied to the `$draft`?

src/Http/ResultTransformerFactory.php
```diff
@@ -4,21 +4,33 @@
...
+            $transformerStack = new CompositeJsonTransformer(
                 new JsonLdEmbeddingJsonTransformer(),
                 new RegionEmbeddingJsonTransformer()
             );
+
+            if (!empty($calendarSummaryFormats)) {
+                $transformerStack = $transformerStack->addTransformer(
+                    new CalendarSummaryEmbeddingJsonTransformer($calendarSummaryFormats)
+                );
+            }
```
bertramakers, line 31: If I'm not mistaken the calendar summaries will now only be embedded if `embed=true` because this is inside `if ($embedded)`. I think it would be more predictable if we can use them independently from each other? So if someone passes `embedCalendarSummaries[]=lg-html` but not explicitly `embed=true` that they still get the calendar summary embedded at least?

src/JsonDocument/CompositeJsonTransformer.php
```diff
@@ -16,6 +16,14 @@ public function __construct(JsonTransformer ...$jsonTransformers)
         $this->jsonTransformers = $jsonTransformers;
     }
 
+    public function addTransformer(JsonTransformer $jsonTransformer): self
+    {
+        $clone = clone $this;
+        $clone->jsonTransformers[] = $jsonTransformer;
+
+        return $clone;
```
bertramakers, line 24: Is this a CS rule that we need a blank line before the return statement? Because I haven't seen this often before in these methods where we make a clone, until recently. But you didn't include it here for example https://github.com/cultuurnet/udb3-search-service/pull/99/files#diff-6f56bca835cff0a172f381493089ea8af2a36e0de6dd3bd957f42660b6e834cdR45

Should we add that as a rule or is it just a preference?

src/Http/ResultTransformerFactory.php
bertramakers: In hindsight I made that `embed` parameter too vague in the past. It would have been better if it was called something like `embedJsonLD` and we'd also have added a separate `embedRegions` when we added that a while back. But if we refactor that now we'd at least have to provide some kind of backward compatibility so maybe not worth it?

src/JsonDocument/CompositeJsonTransformer.php
jdrieghe: There's no rule configured for this. I don't have a strong preference either way.

src/Http/ResultTransformerFactory.php
jdrieghe: Fixed that.

src/ElasticSearch/JsonDocument/CalendarSummaryEmbeddingJsonTransformer.php
jdrieghe: Yeah, that was super confusing because on the outside it looks like the `$draft` argument isn't used until you look at the reducer in `CompositeJsonTransformer` 😅

LucWollants: Maybe `source` and `destination` are better names? (No need to change, the interface describes it in a comment)

src/JsonDocument/CompositeJsonTransformer.php
LucWollants: I think this is an example of something that's up to the developer to decide. I guess a bit of freedom for personal taste should still be allowed 😀

LucWollants (approved):

jdrieghe:
Let's make sure to give this a proper manual test before merging though.

LucWollants:
> Let's make sure to give this a proper manual test before merging though.

I can do that on Monday if you want.

src/JsonDocument/CompositeJsonTransformer.php
bertramakers: All good 👍

jdrieghe:
@LucWollants let's do that together then! 🎉

bertramakers (approved):
Code looks great! 👍 Let's do a manual test and then merge it.

PR #102, opened 2021-03-08
III-3840 Min and max values for start and limit

bertramakers:
### Added
 
- Added min and max value on `limit` parameter on `/offers/`, `/events/`, `/places/`, and `/organizers/`
- Added min and max value on `start` parameter on `/offers/`, `/events/`, `/places/`, and `/organizers/`
 
### Removed
 
- Removed usages of `Natural` for start and limit parameters, and in `PagedResultSet` class
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3840

I based the values on the ones documented here: https://documentatie.uitdatabank.be/content/search_api_3/latest/getting-started/pagination.html

limit=2000 was around the threshold that made PHP run out of memory in the past I think.
start=10000 is based on the max starting position supported by Elasticsearch

### DIFF
--- a/src/ElasticSearch/AbstractElasticSearchQueryBuilder.php
+++ b/src/ElasticSearch/AbstractElasticSearchQueryBuilder.php
@@ -97,10 +97,10 @@ abstract class AbstractElasticSearchQueryBuilder implements QueryBuilder
         return $c;
     }
 
-    public function getLimit(): Natural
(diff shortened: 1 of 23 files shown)


jdrieghe (approved):
Nice! Loving the extra campground cleaning here.

bertramakers:
Thanks! 😃

PR #106, opened 2021-03-10
III-3851 Refactor Country value object

LucWollants:
### Added
- Added custom `Country` value object which no longer requires a separate `CountryCode` class
---


src/Country.php
```diff
@@ -0,0 +1,290 @@
...
+
+use InvalidArgumentException;
+use ReflectionClass;
+
+final class Country
+{
+    public const AF = 'AF';
+    public const AX = 'AX';
+    public const AL = 'AL';
+    public const DZ = 'DZ';
```
bertramakers, line 15: I'm not sure if it's useful to be this strict and only allow "known" country codes, since we would have to keep this list up-to-date. Maybe we can just make it check that the string is 2 uppercase characters from A-Z to make this simpler? `AA` would be accepted as valid while it's technically not a country code, but I'm not sure if we need to go that far in our validation.

bertramakers (approved):

src/Country.php
bertramakers: I was looking in the alternative if only having a list of only `BE` and `NL` right now, but that doesn't seem right because there's ~2500 events and places in other countries apparently: https://io.uitdatabank.be/offers/?apiKey=0a5a9ce4-7066-11e7-a783-0271c1cbe993&disableDefaultFilters=true&q=_exists_:address.\*.addressCountry%20AND%20NOT(address.\*.addressCountry:BE%20OR%20address.\*.addressCountry:NL)

For example this event in the US: https://io.uitdatabank.be/event/f2bbdd1d-b63c-4751-8adc-9f914e2599bf

Other use cases are sports camps organized in Belgium but happening in other countries.

And that would also be a list that we would need to keep up-to-date whenever we want to add support for another country.

LucWollants: I think the list of countries doesn't change that often so that the added value of being strict is worth the cost to maintain this list?

bertramakers: I don't see it as much as a cost in the actual work to add/remove/change country codes, but more that we have to make sure that we follow up on it. Even if they don't change often, we still need to keep track of when they change. Stuff like that compounds quickly I've noticed, we also have a hard time keeping the list of Belgian postal codes in sync and up-to-date across all of our products. 

And when you then have data from one system going into another where the list of options is outdated, it usually takes some time to debug before it actually becomes clear that the problem is an outdated list of values. (Data just appears to be missing -> Look in logs -> Messages get discarded, why? -> etc)

Plus we don't even know if the list here is correct, assuming you copied it from the cultuurnet/valueobjects library which was a fork from one that is from 2013.

Granted, maybe I'm looking too much into this since it worked fine before when we were using cultuurnet/valueobjects. But now feels like a good time to re-evaluate how these VOs should behave now that we internalize them and can determine for ourselves how they work in a given context? Because for example, I'm wondering what the added value of being strict exactly is here? Since the data being indexed comes from UDB3, so it's already persisted in the single source of truth (= event store) there before it gets indexed here. Worst case, I think, someone does something like `?addressCountry=AA` and they don't get any results. Unless you see other risks?

LucWollants: It is indeed for read-actions and not for write-actions. So I will take the 2 character approach.

bertramakers (approved):

PR #107, opened 2021-03-10
III-3851 Fix time type

LucWollants:
### Added
- Added new `Timestamp` class
---


src/ElasticSearch/Offer/ElasticSearchOfferQueryBuilder.php
```diff
@@ -154,7 +155,7 @@ public function withDateRangeFilter(
 
     public function withLocalTimeRangeFilter(int $localTimeFrom = null, int $localTimeTo = null): self
     {
-        $this->guardNaturalIntegerRange('localTime', new Age($localTimeFrom), new Age($localTimeTo));
+        $this->guardNaturalIntegerRange('localTime', new Timestamp($localTimeFrom), new Timestamp($localTimeTo));
```
bertramakers, line 158: Nit: This is technically not a timestamp in the traditional sense of the word, but an int between `0` and `2359`. Maybe we can add that constraint to the VO and rename it to just `Time`?

LucWollants: Yes!

LucWollants: Changed name to `Timestamp` and added range check between `0` and `2359`

bertramakers (approved):

PR #108, opened 2021-03-10
III-3851 Remove dependency on Enums

LucWollants:
### Changed
- Refactor `SortOrder` to custom enum implementation
- Refactor `FallbackType` to custom enum implementation
- Refactor `FacetName` to custom enum implementation
- Refactor `OfferType` to custom enum implementation
- Refactor `Status` to custom enum implementation
---


src/ElasticSearch/JsonDocument/Properties/FallbackType.php
```diff
@@ -4,16 +4,56 @@
...
+    /**
+     * @var string
+     */
+    private $fallbackType;
+
+    public function __construct(string $fallbackType)
+    {
+        if (!\in_array($fallbackType, $this->getAllowedValues())) {
+            throw new InvalidArgumentException(
+                'The given fallback type ' . $fallbackType . ' is not Event, Place or Organizer'
```
jdrieghe, line 24: I often use something like this:
```
'Invalid FallbackType: ' . $fallbackType . '. Should be one of join(',', self::ALLOWED_VALUES)'
```

src/ElasticSearch/JsonDocument/Properties/FallbackType.php
```diff
@@ -4,16 +4,56 @@
...
+    {
+        return new self(self::ORGANIZER);
+    }
+
+    public function toString(): string
+    {
+        return $this->fallbackType;
+    }
+
+    private function getAllowedValues(): array
```
jdrieghe, line 51: This can be a `private const`

src/ElasticSearch/JsonDocument/Properties/FallbackType.php
```diff
@@ -4,16 +4,56 @@
...
- * @method static FallbackType ORGANIZER()
- */
-final class FallbackType extends Enum
+use InvalidArgumentException;
+
+final class FallbackType
 {
     public const EVENT = 'Event';
     public const PLACE = 'Place';
     public const ORGANIZER = 'Organizer';
```
jdrieghe, line 13: Do we need these internals to be public?

jdrieghe (changes requested):
I left some questions/suggestions on the FallbackType enum, but the same applies to all the others

src/ElasticSearch/JsonDocument/Properties/FallbackType.php
LucWollants: That is indeed a much better error message! Will modify the enums.

src/ElasticSearch/JsonDocument/Properties/FallbackType.php
LucWollants: I even think the method is not needed. I can do something like this
```
final class FallbackType
{
    private const EVENT = 'Event';
    private const PLACE = 'Place';
    private const ORGANIZER = 'Organizer';

    private static $allowedValues = [
        self::EVENT,
        self::PLACE,
        self::ORGANIZER,
    ];
```

src/ElasticSearch/JsonDocument/Properties/FallbackType.php
LucWollants: They can be indeed private!

LucWollants:
> I left some questions/suggestions on the FallbackType enum, but the same applies to all the others

@jdrieghe thanks for those remarks. I will tweak all custom Enum classes.

src/ElasticSearch/JsonDocument/Properties/FallbackType.php
jdrieghe: I'd even make it `private const ALLOWED_VALUES` to make sure nothing can change it

jdrieghe (approved):

PR #110, opened 2021-03-10
III-3581 Replace UUID from value objects with Uuid from Ramsey

LucWollants:
### Changed
- Replace UUID from value objects with Uuid from Ramsey
---
Ticket: https://jira.uitdatabank.be/browse/III-3581


src/AMQP/EventBusForwardingConsumer.php
```diff
@@ -11,7 +11,7 @@
 use Broadway\EventHandling\EventBusInterface;
 use CultuurNet\UDB3\Search\Deserializer\DeserializerLocatorInterface;
 use PhpAmqpLib\Connection\AMQPStreamConnection;
-use ValueObjects\Identity\UUID;
+use Rhumsaa\Uuid\Uuid;
```
jdrieghe, line 14: This is still using the old (abandone) namespace. Can we update this dependency to a more recent version?

LucWollants: Ouch! Will do!

LucWollants: To fix this I also had to upgrade `broadway/broadway` and `cultuurnet/valueobjects`.

After that upgrade, I did a manual test and changed the postal code of a location.
Everything got re-indexed to the new postal code and searching required the new postal code.

jdrieghe (approved):

PR #121, opened 2021-03-17
Make symfony/console a regular dependency and not just a dev dependency

bertramakers:
### Fixed
 
- Fixed broken `./bin/app.php` script when using `composer install --no-dev`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         "ramsey/uuid": "^3.9",
         "sentry/sdk": "^2.2",
         "slim/psr7": "^0.6",
+        "symfony/console": "^4.4.20",
         "symfony/finder": "~3.2",
         "symfony/yaml": "^4.3",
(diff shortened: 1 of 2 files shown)


LucWollants (approved):

composer.json
```diff
@@ -32,6 +32,7 @@
         "ramsey/uuid": "^3.9",
         "sentry/sdk": "^2.2",
         "slim/psr7": "^0.6",
+        "symfony/console": "^4.4.20",
```
jdrieghe, line 35: Why is this set specifically to `^4.4.20`?
This way it will never update to a potential `4.5`

bertramakers: Ah yes. I just took the version we had in the lock file.

bertramakers: https://github.com/cultuurnet/udb3-search-service/pull/122

LucWollants: I will also try to remember this 😊

PR #125, opened 2021-03-17
Remove stringy dependency

bertramakers:
### Removed
 
- Removed stringy dependency


LucWollants (approved):
Another one bites the dust

PR #126, opened 2021-03-17
III-3816 Consistent error/exception logging

bertramakers:
### Added

- Added `/bin` to php-cs-fixer config
- Added some exception types to better identify user-caused exceptions

### Changed

- Almost all HTTP exceptions now get logged to a `/log/web.log` file (as well as Sentry), except for common user exceptions
- Almost all CLI exceptions now get logged to a `/log/cli.log` file (as well as Sentry), except for common user exceptions
- Sentry integration now happens via Sentry's Monolog handler
- Simplified logger instantiation of other log files

---
Ticket: https://jira.uitdatabank.be/browse/III-3816

After this is merged and deployed to acc/test/prod, we should be able to clean up the Sentry errors for this service and start actually following up on them.


app/LoggerProvider.php
```diff
@@ -4,41 +4,25 @@
...
-                    Logger::DEBUG
-                );
-                $logger->pushHandler($logFileHandler);
-
-                return new SentryPsrLoggerDecorator(
-                    $this->get(SentryExceptionHandler::class),
-                    $logger
+                return LoggerFactory::create(
+                    $this->getContainer(),
+                    new LoggerName('amqp', 'amqp.udb3_publisher'),
```
LucWollants, line 24: I am confused by the naming, we register a `logger.amqp.udb3_consumer` and use `amqp.udb3_publisher` as file name? (It has been like that from the start)

bertramakers: - `logger.amqp.udb3_consumer` is the name in the service container
- `amqp(.log)` is the file name
- `amqp.udb3_publisher` is the logger name, which gets sent to Sentry

I didn't change any of these because I figured changing the implementation would be more important to get right first, so our logging works the same consistently. I'm open to renaming files and logger names but we might need to check with @willaerk if there's an impact on the log rotation and maybe some logs get parsed etc. So I would split that off to another ticket/PR for later then?

bertramakers: In general I think it would be good to clean up the log(ger) names sometime, as they are very inconsistent and sometimes unclear: https://github.com/cultuurnet/udb3-silex#logs (e.g. `uitpas-events.log` + `uitpas.log` -> Super unclear which is for what)

I would love to for example clarify them, group them with prefixes like e.g. `amqp.<log>.log` etc. But it doesn't feel high priority I think and might require too much coordination than what it's worth right now?

src/Http/Hydra/PagedCollection.php
```diff
@@ -42,11 +42,6 @@ public function __construct(
 
     private function setPageNumber(int $pageNumber): void
     {
-        if (!is_int($pageNumber)) {
```
LucWollants, line 45: Primitive type hinting can be so useful!

bertramakers: Yes I was surprised how many checks we still had for that 😄

app/LoggerProvider.php
LucWollants: So I am correct to assume that `udb3_consumer` is more logic then `udb3_publisher`? If so I think we should create a new ticket to make this consistent.

LucWollants (approved):
Great work!

app/LoggerProvider.php
bertramakers: Yes you are correct. I think we should create a ticket to review all the logger names in general and make them more consistent and predictable.

bertramakers: https://jira.uitdatabank.be/browse/III-3881

PR #142, opened 2021-04-12
III-3951 jsonld polyfill transformer

bertramakers:
### Added
 
- Added polyfill transformer similar to the one in UDB3 to correct some malformed properties at run-time (since the JSON-LD is cached in SAPI3 to prevent tens/hundreds of HTTP requests to udb3 which incur too much of a performance hit)
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3951


jdrieghe (approved):

bertramakers:
Thanks for merging this, I forgot about it! 😄

PR #144, opened 2021-04-19
III-3934 - Use index 'no' instead of 'not-analyzed' to bypass the 32kb limit

jdrieghe:
### Fixed
- Some events with a very large JSON-LD representation did not get indexed properly
 
---
Ticket: https://jira.uitdatabank.be/browse/III-3934

### DIFF
--- a/src/ElasticSearch/Operations/SchemaVersions.php
+++ b/src/ElasticSearch/Operations/SchemaVersions.php
@@ -6,6 +6,6 @@ namespace CultuurNet\UDB3\Search\ElasticSearch\Operations;
 
 final class SchemaVersions
 {
-    public const UDB3_CORE = 20210318153800;
+    public const UDB3_CORE = 20210419100800;
     public const GEOSHAPES = 20190111135400;
 }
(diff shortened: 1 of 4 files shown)


LucWollants (approved):
Will this require a full replay?

jdrieghe:
I think it's mostly to solve new documents already, but indeed. A re-index would not hurt I think.

bertramakers:
Thanks for picking this up! You will also have to increment the schema version, otherwise the change won't be picked up even when we re-index (at least locally where we use an idempotent command that checks the version and only creates a new index if the version is different - maybe Kristof does it differently on acc/test/prod): https://github.com/cultuurnet/udb3-search-service/blob/main/src/ElasticSearch/Operations/SchemaVersions.php

bertramakers:
And yes we will need a re-index, because the new mapping will only take effect when we create a new index with this mapping.

jdrieghe:
@bertramakers done

bertramakers (approved):

PR #156, opened 2021-06-10
III-4047 Simplify authentication middleware

LucWollants:
### Added
 
- Added custom `ApiProblem` classes for apiKey authentication
- Added `Consumer` value object
- Added new `AuthenticateRequest` middleware without api-guard dependency

### Removed
- Removed the `AuthenticateRequest` with the `cultuurnet/udb3-api-guard` dependency
- Removed the `cultuurnet/udb3-api-guard` dependency

### Changed
- Updated `OrganizerSearchController` to work with `Consumer`
- Updated `OfferSearchController` to work with a `Consumer`
- Use `Consumer` instead of `ApiKey` inside Sentry Handler
---

Ticket: https://jira.uitdatabank.be/browse/III-4047


src/Http/Authentication/ApiProblems/BlockedApiKey.php
```diff
@@ -0,0 +1,27 @@
...
+use CultuurNet\UDB3\Search\Http\ResponseFactory;
+use Psr\Http\Message\ResponseInterface;
+
+final class BlockedApiKey extends ApiProblem
+{
+    public function __construct(string $apiKey)
+    {
+        parent::__construct('Unauthorized', 'https://api.publiq.be/probs/auth/unauthorized');
+        $this->setStatus(401);
+        $this->setDetail('the provided api key ' . $apiKey . ' is blocked');
```
bertramakers, line 17: Nit: Maybe we should capitalize the start of the detail sentence? I.e. `The provided api key ...`?

bertramakers: I see you did it for `No x-api-key header or apiKey parameter found` down below, but not for the ones that start with `The provided api key ...`

app/Error/SentryHandlerScopeDecorator.php
```diff
@@ -55,7 +54,7 @@ public function handle(array $record): bool
         withScope(function (Scope $scope) use ($record, &$result): void {
             $scope->setTags(
                 [
-                    'api_key' => $this->apiKey->toString(),
+                    'id' => $this->consumer->getId(),
```
bertramakers: Just a minor suggestion: Maybe we should also add some kind of "consumer type" to the `Consumer` class, so that we can add it as a tag here. Possible values could for example be `uitidv1` (in this PR) and `auth0` (in a later PR). (Or just `uitid` instead of `uitidv1`, although we sometimes also use `uitidv2` as name for `auth0` internally 🤔)

bertramakers:
Very thorough PR, and the commits were super easy to follow along! 👏

bertramakers (approved):

app/Error/SentryHandlerScopeDecorator.php
bertramakers: (I'm going to check what "uitid v2" exactly means internally)

LucWollants: Yes, some extra "context" can be useful. Maybe we do that with the follow up [ticket](https://jira.uitdatabank.be/browse/III-4048) when we add clientId?

bertramakers: Sounds good!

PR #163, opened 2021-06-14
III-4048 Support clientId for Auth0 integration

LucWollants:
### Added
- Added `auth0/auth0-php` dependency
### Changed
- Extend `AuthenticateRequest` middleware with support for clientId
---
Ticket: https://jira.uitdatabank.be/browse/III-4048

### DIFF
--- a/src/Http/Authentication/ApiProblems/MissingSapiScope.php
+++ b/src/Http/Authentication/ApiProblems/MissingSapiScope.php
@@ -12,8 +12,8 @@ final class MissingSapiScope extends ApiProblem
 {
     public function __construct(string $clientId)
     {
-        parent::__construct('Unauthorized', 'https://api.publiq.be/probs/auth/unauthorized');
-        $this->setStatus(401);
+        parent::__construct('Forbidden', 'https://api.publiq.be/probs/auth/forbidden');
+        $this->setStatus(403);
         $this->setDetail('The provided client id ' . $clientId . ' has missing sapi3 scope');
     }
 
--- a/src/Http/Authentication/ApiProblems/MissingSapiScope.php
+++ b/src/Http/Authentication/ApiProblems/MissingSapiScope.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search\Http\Authentication\ApiProblems;
+
+use Crell\ApiProblem\ApiProblem;
+use CultuurNet\UDB3\Search\Http\ResponseFactory;
+use Psr\Http\Message\ResponseInterface;
+
+final class MissingSapiScope extends ApiProblem
+{
+    public function __construct(string $clientId)
+    {
+        parent::__construct('Unauthorized', 'https://api.publiq.be/probs/auth/unauthorized');
+        $this->setStatus(401);
+        $this->setDetail('The provided client id ' . $clientId . ' has missing sapi3 scope');
+    }
+
+    public function toResponse(): ResponseInterface
+    {
+        return ResponseFactory::jsonLd(
+            $this->asArray(),
+            $this->getStatus()
+        );
+    }
+}
--- a/src/Http/Authentication/Auth0TokenGenerator.php
+++ b/src/Http/Authentication/Auth0TokenGenerator.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search\Http\Authentication;
+
+use GuzzleHttp\Client;
+
+final class Auth0TokenGenerator
+{
+    /**
+     * @var Client
+     */
+    private $client;
+
+    /**
+     * @var string
+     */
+    private $domain;
+
+    /**
+     * @var string
+     */
+    private $clientId;
+
+    /**
+     * @var string
+     */
+    private $clientSecret;
+
+    public function __construct(Client $client, string $domain, string $clientId, string $clientSecret)
+    {
+        $this->domain = $domain;
+        $this->client = $client;
+        $this->clientId = $clientId;
+        $this->clientSecret = $clientSecret;
+    }
+
+    public function newToken(): string
+    {
+        $response = $this->client->post(
+            'https://' . $this->domain . '/oauth/token',
+            [
+                'headers' => ['content-type' => 'application/x-www-form-urlencoded'],
+                'body' => sprintf(
+                    'grant_type=client_credentials&client_id=%s&client_secret=%s&audience=%s',
+                    $this->clientId,
+                    $this->clientSecret,
+                    'https://' . $this->domain . '/api/v2/'
+                ),
+            ]
+        );
+
+        $res = json_decode($response->getBody()->getContents(), true);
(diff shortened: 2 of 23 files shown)


src/Http/Authentication/Auth0TokenGenerator.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->clientId = $clientId;
+        $this->clientSecret = $clientSecret;
+    }
+
+    public function newToken(): string
+    {
+        $response = $this->client->post(
+            'https://' . $this->domain . '/oauth/token',
+            [
+                'headers' => ['content-type' => 'application/x-www-form-urlencoded'],
```
bertramakers: Nit: You should also be able to do this with a JSON body. That's maybe easier?

bertramakers: You can see an example on https://publiq.stoplight.io/docs/authentication/docs/client-access-token.md
The tokens you need to communicate with the management API are the same as the machine-to-machine tokens for clients that want to access our APIs. The only difference is that the client you're using here has access to the management API while other clients do not.
(Ignore the part about scopes, that is not relevant and will also be removed soon.)

LucWollants: That also works and is indeed more readable and a better fit the kind of Auth0 consumer SAPI3 is.

src/Http/Authentication/AuthenticateRequest.php
```diff
@@ -41,12 +49,42 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
         }
 
+        return $this->handleApiKey($request, $handler, $apiKey);
+    }
+
+    private function handleClientId(ServerRequestInterface $request, RequestHandlerInterface $handler, string $clientId): ResponseInterface
+    {
+        try {
+            $metadata = $this->auth0Client->getMetadata($clientId, $this->auth0Client->getToken());
+        } catch (Exception $exception) {
```
bertramakers, line 71: I remember that we said not to reject a client id when Auth0 is down. Is there a way to differentiate exceptions here between "the client does not exist" and "auth0 is down"?

bertramakers: I would assume that if auth0 is down we don't get a 4xx error, but either a 5xx or no real response at all (not sure how Guzzle handles that)

LucWollants: I will try to simulate this with an invalid endpoint or url.

src/Http/Authentication/AuthenticateRequest.php
```diff
@@ -41,12 +50,42 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+
+    private function handleClientId(ServerRequestInterface $request, RequestHandlerInterface $handler, string $clientId): ResponseInterface
+    {
+        try {
+            $client = $this->auth0Management->clients()->get($clientId, ['client_id', 'client_metadata'], true);
+        } catch (Exception $exception) {
+            return (new InvalidClientId($clientId))->toResponse();
+        }
+
+        if (empty($client['client_metadata']['sapi3'])) {
```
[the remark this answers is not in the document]
LucWollants: If we keep the "true/false" approach I will also do a typecast to boolean.
If we take the separated string approach I will adapt to that approach.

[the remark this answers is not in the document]
bertramakers: We are going with the separated string approach, because the `true/false` approach would make us run into the limit of max 10 properties inside the metadata quickly. (Especially because we also use 2 other properties already in some clients, and might need some more in other clients)

src/Http/Authentication/AuthenticateRequest.php
bertramakers: Since it can be hard to predict what we will get back when Auth0 is actually down (vs an invalid endpoint), we could maybe only consider the client id invalid if the response code is in the 4xx range? Since that indicates specifically that there is a problem with the client id, instead of a problem with Auth0 itself.

src/Http/Authentication/ApiProblems/MissingSapiScope.php
```diff
@@ -0,0 +1,27 @@
...
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search\Http\Authentication\ApiProblems;
+
+use Crell\ApiProblem\ApiProblem;
+use CultuurNet\UDB3\Search\Http\ResponseFactory;
+use Psr\Http\Message\ResponseInterface;
+
+final class MissingSapiScope extends ApiProblem
```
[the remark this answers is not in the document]
LucWollants: Renamed to `NotAllowedToUseSapi`

src/Http/Authentication/AuthenticateRequest.php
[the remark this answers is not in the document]
LucWollants: Used the new `public-apis` list

src/Http/Authentication/AuthenticateRequest.php
LucWollants: I tried with an invalid domain and `ConnectionException` seems to be the desired exception to implement this bypass of the validation

bertramakers: Is that also for responses that have a 5xx status?
Because I think we need to account for either Auth0 not being reachable at all (connection exceptions), and Auth0 returning internal errors (500, 503, ...)

bertramakers: Oh and maybe timeout exceptions are also another kind?

LucWollants: We could also take another approach and add a configuration to disable apiKey and/or clientId validation?

bertramakers: That already exists I think? But it requires manual intervention then (both to disable and re-enable). I think that we should be good if we disable the automatic exception throwing from Guzzle, and then check the response code ourselves. If it's 404 we return a Forbidden (client does not exist), if it's 200 we check the metadata like we do now, and if it's anything else we temporarily allow it? And we should then still have a try/catch around it for connection exceptions I think, because a timeout for example doesn't return a response I imagine.

LucWollants: I proposed the manual intervention to keep the code simple and avoid possible edge-cases were we would always allow access. So in case Auth0 goes down we have a "emergency plan" and disable the `Auth0` check inside `udb3-search`?

LucWollants: @bertramakers I reworked the code to closely match your proposal:
- I disabled Guzzle exceptions
- Everything else then a 200 response code returned from Auth0 is treated as missing metadata and results in an `InvalidClientId`
- A `ConnectionException` is still handled and is equivalent to Auth0 being down
- Other possible exceptions are ignored and bubble up to make sure coding/logical errors are not missed

LucWollants: @bertramakers as discussed the 500 HTTP errors are handled as ConnectionException

PR #164, opened 2021-06-15
III-4057 Add cache for Auth0 token

LucWollants:
### Added
- Added file based cached for Auth0 token`Auth0TokenFileRepository` 
- Added provider `Auth0TokenProvider` for an Auth0 token either from cache or new from Auth0
### Changed
- Refactor AuthenticateRequest to use Auth0TokenProvider
---
Ticket: https://jira.uitdatabank.be/browse/III-4057

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,4 @@
 /web/autocomplete.json
 /.php_cs.cache
 /.php-cs-fixer.cache
-/.auth0-token-cache
+/.auth0-token-cache.json
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /web/autocomplete.json
 /.php_cs.cache
 /.php-cs-fixer.cache
+/.auth0-token-cache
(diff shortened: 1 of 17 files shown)


LucWollants:
@bertramakers I added an option to pass an "offset" to the expires method so it's possible to check if a token would expire in 5 minutes. If the token would expire within 5 minutes it's already replaced from the cache.

bertramakers:
> @bertramakers I added an option to pass an "offset" to the expires method so it's possible to check if a token would expire in 5 minutes. If the token would expire within 5 minutes it's already replaced from the cache.

We're still parsing the token to check the expiration time now though. Ideally we'd use the `expires_in` that's included in the response when we request the token, because Auth0 discourages parsing the token (I think because they do not guarantee that it will always be a JWT, it can also be an opaque token in some cases).

LucWollants:
@bertramakers I took some time to extend the code with a simple `Auth0Token` value object. This `Auth0Token` is created from the token string and the expires in values returned from Auth0. Add `Auth0Token` construction time the current date time is also stored as the issues at value.

This new `Auth0Token` is then stored inside a simple file cache with a JSON format. When the token would expire with 5 minutes the cache is cleared and a new token is asked.

bertramakers (approved):
Nice work, I like the approach with the Auth0Token VO!

PR #166, opened 2021-06-28
III-3934 Replace "no" with false

bertramakers:
### Changed
 
- Set `enabled` for `originalEncodedJsonLd` to `false`
- Set `type` of `originalEncodedJsonLd` to `object` so it does not get parsed as some kind of string

---

Ticket: https://jira.uitdatabank.be/browse/III-3934

The documentation for `enabled` for Elasticsearch 5.3: https://discuss.elastic.co/t/why-elasticsearch-still-check-the-length-of-a-keyword-field-even-if-its-not-indexed/134604/5

In more recent documentation it is clarified that you can also set the type to `object` for string values if you set `enable` to `false`: https://www.elastic.co/guide/en/elasticsearch/reference/7.x/enabled.html

> Note that because Elasticsearch completely skips parsing the field contents, it is possible to add non-object data to a disabled field

I got this solution from this question which is exactly the scenario that we were in: https://discuss.elastic.co/t/why-elasticsearch-still-check-the-length-of-a-keyword-field-even-if-its-not-indexed/134604/5

Tested this locally with a huge event that I imported, and I was able to reproduce the bug and also verify that this fixed it.

### DIFF
--- a/src/ElasticSearch/Operations/SchemaVersions.php
+++ b/src/ElasticSearch/Operations/SchemaVersions.php
@@ -6,6 +6,6 @@ namespace CultuurNet\UDB3\Search\ElasticSearch\Operations;
 
 final class SchemaVersions
 {
-    public const UDB3_CORE = 20210628134100;
+    public const UDB3_CORE = 20210629145300;
     public const GEOSHAPES = 20190111135400;
 }
--- a/src/ElasticSearch/Operations/SchemaVersions.php
+++ b/src/ElasticSearch/Operations/SchemaVersions.php
@@ -6,6 +6,6 @@ namespace CultuurNet\UDB3\Search\ElasticSearch\Operations;
 
 final class SchemaVersions
 {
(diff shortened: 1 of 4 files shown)


LucWollants (approved):
The change looks fine, but how is this related to https://jira.uitdatabank.be/browse/III-4026

bertramakers:
> The change looks fine, but how is this related to https://jira.uitdatabank.be/browse/III-4026

The `index: "no"` was an attempt to fix that out of memory error by not indexing this large property (we don't need to filter/query on it anyway), but it didn't work.

bertramakers:
Ah sorry wrong ticket number, it's this one: https://jira.uitdatabank.be/browse/III-3934

bertramakers:
I've tested this locally with an event I imported with a lot of subevents and it still does not get indexed and I still get the error locally, so I will need to try something else tomorrow first.

LucWollants (approved):

PR #188, opened 2021-08-09
III-4033 Implement searching on booking availability

LucWollants:
### Added
- Added booking availability to mapping of place and event
- Set booking availability on top level of place or event
- Set booking availability on sub events of place or event
- Added booking availability as query param

### Changed
- Changed GitHub actions to 7.4 exclusive

### Removed
- Removed setting status through polyfill

---
Ticket: https://jira.uitdatabank.be/browse/III-4033


src/Offer/BookingAvailability.php
```diff
@@ -0,0 +1,51 @@
...
+     */
+    private const ALLOWED_VALUES = [
+        self::AVAILABLE,
+        self::UNAVAILABLE,
+    ];
+
+    private function __construct(string $value)
+    {
+        if (!\in_array($value, self::ALLOWED_VALUES, true)) {
+            throw new InvalidArgumentException('Booking availability does not support the value "' . $value . '"');
```
bertramakers, line 27: Okay for now but something to discuss at some point in the larger picture of (request) validation: I'm not sure of the added value of VO's in SAPI3 for the query parameters. 

1) We don't use them to validate data that gets persisted, only to query data
2) If we move to request validation using JSON schema and OpenAPI files like we discussed on Slack, then we can document the possible values in those and have them validated in the request handler / controller and after that it doesn't really matter
3) If we add a possible value we have to remember to add it here as well
4) This throws `InvalidArgumentException` which will be hard to convert to an `ApiProblem` with the correct type once we clean that up everywhere, since `InvalidArgumentException` is so vague

Most other VOs don't even have any validation as far as I can see in SAPI3, except for `Status`. For example `CalendarType` and `WorkflowStatus` don't.

And the more I think about it, the less I see the need for these VOs in general in SAPI3, because the only data that we persist is coming from UDB3 and if it's a specific value in UDB3 we just have to accept that here. And if it's a query parameter, worst case that happens is that you don't get any results if you use an unknown value. (Which again we can also prevent using the request validation on a higher level).

bertramakers: Instead of `InvalidArgumentException`, we should probably throw this one instead though: https://github.com/cultuurnet/udb3-search-service/blob/main/src/Offer/Status.php#L7

So at least that can get converted to the correct ApiProblem later.

LucWollants: Indeed there is no added value in using a VO here. If the value provided for `bookingAvailability` would be something that is not supported the result set will just be empty.

So it's easy enough to remove `BookingAvailability` and just use a string for it.

bertramakers:
Nice work @LucWollants 👏

PR #191, opened 2021-09-09
III-4214 Remove QSA flag

bertramakers:
### Removed
 
- Removed `QSA` flag in rewrite rules. This would combine the original query string with the query string in our rewrite rule, and causes the query string parameters to be reordered. However there is no query string in our new URL, so there is nothing to combine it with and it just reorders the original query parameters.
 
---

Ticket: https://jira.uitdatabank.be/browse/III-4214

### DIFF
--- a/web/.htaccess
+++ b/web/.htaccess
@@ -7,5 +7,5 @@
     RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
 
     RewriteCond %{REQUEST_FILENAME} !-f
-    RewriteRule ^(.*)$ index.php [QSA,L]
+    RewriteRule ^(.*)$ index.php [L]
 </IfModule>


jdrieghe (approved):
Will this hurt us when we introduce endpoints with query string parameters in the future? Is that likely?

bertramakers:
Well our endpoints on SAPI3 (this app) do use query string parameters. But even without the `QSA` flag they get appended. They would only get discarded if the rewrite URL (`index.php`) had query parameters, but it doesn't and will probably never in the future either. We just want to append the whole original URL (for example `/events/?sort[score]=desc` to `index.php`, so it becomes `index.php/events/?sort[score]=desc`). And this seems to work fine without the `QSA` flag.

When we do use the QSA flag the order of the query parameters changes sometimes and this breaks sorting on multiple fields (see the ticket).

willaerk:
This is only for the search service, and only applies when the path you are rewriting to (in casu /index.php) would contain parameters. Then the QSA flag merges the parameters from the request with the parameters from the URL you rewrite to.

In our situation, I think the /index.php path will never have parameters, as it is only the entry point to the application.
So it is safe to remove.

PR #199, opened 2021-11-24
III-4386 Fix recommendation sorting

LucWollants:
### Fixed
- Fixed recommendation sorting by using nested type
 
---

Ticket: https://jira.uitdatabank.be/browse/III-4386

### DIFF
--- a/tests/Http/OfferSearchControllerTest.php
+++ b/tests/Http/OfferSearchControllerTest.php
@@ -1050,6 +1050,25 @@ final class OfferSearchControllerTest extends TestCase
     /**
      * @test
      */
+    public function it_throws_an_exception_for_sort_order_recommendation_score_and_missing_recommendation_for(): void
+    {
+        $request = $this->getSearchRequestWithQueryParameters(
+            [
+                'sort' => [
+                    'recommendationScore' => 'asc',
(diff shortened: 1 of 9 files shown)


tests/Http/OfferSearchControllerTest.php
```diff
@@ -1047,6 +1047,25 @@ public function it_throws_an_exception_for_sort_order_distance_and_missing_coord
...
+    {
+        $request = $this->getSearchRequestWithQueryParameters(
+            [
+                'sort' => [
+                    'recommendationScore' => 'asc',
+                ],
+            ]
+        );
+
+        $this->expectException(\InvalidArgumentException::class);
```
bertramakers, line 1063: Nit: It's technically a `MissingParameter`, which indeed extends from `InvalidArgumentException`. But you could more specifically expect `MissingParameter` here to be safe. (Better to have these specific exceptions so we can add better support for API problem responses with specific types later)

bertramakers (approved):

tests/Http/OfferSearchControllerTest.php
LucWollants: Good catch! I changed it.

PR #200, opened 2022-01-25
III-4424 Index organizer region

LucWollants:
### Added
- Added geo related fields to organizer mapping and incremented schema
- Added factory `GeoDistanceParametersFactory` to create `GeoDistanceParameters` from `ApiRequestInterface`
- Added factory `GeoBoundsParametersFactory` to create `GeoBoundsParameters` from `ApiRequestInterface`
- Added `GeoBoundsOrganizerRequestParser` for `Organizer`
- Added `DistanceOrganizerRequestParser` for `Organizer`

### Changed
- Renamed `OfferRegionServiceInterface` to `RegionServiceInterface`
- Renamed `GeoShapeQueryOfferRegionService` to `GeoShapeQueryRegionService`
- Used `GeoDistanceParametersFactory` inside `GeoDistanceOfferRequestParser`
- Used `GeoBoundsParametersFactory` inside `GeoBoundsOfferRequestParser`
- Extended `OrganizerTransformer` with `GeoInformationTransformer`

### Removed
- Removed `OfferType` from `GeoInformationTransformer`

---

Note: the `region` is not yet supported as a URL parameter


app/ElasticSearchProvider.php
```diff
@@ -46,9 +46,9 @@ function () {
         );
 
         $this->add(
-            'offer_region_service',
+            'region_service',
```
bertramakers, line 49: Nit: Maybe use the class name or interface name instead if you rename this anyway

bertramakers (approved):

app/ElasticSearchProvider.php
LucWollants: Good point, I changed it.

PR #202, opened 2022-01-26
III-4424 Improve json encoding

LucWollants:
### Added
- Added `Json` class to have a more consistent way of encoding and decoding to Json
 
### Changed
- Replaced occurrences of `json_encode` with `Json::encode`
- Replaced occurrences of `json_decode` with `Json::decode`
- Boy scouting on the changed classes

---

### DIFF
--- a/src/Http/JsonLdResponse.php
+++ b/src/Http/JsonLdResponse.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Search\Http;
 
+use CultuurNet\UDB3\Search\Json;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\StreamInterface;
 
@@ -30,7 +31,7 @@ final class JsonLdResponse implements ResponseInterface
     public function withData($data)
     {
         $body = $this->response->getBody();
-        $body->write(json_encode($data, self::JSON_OPTIONS));
+        $body->write(Json::encodeWithOptions($data, self::JSON_OPTIONS));
 
         return $this->response->withBody($body);
     }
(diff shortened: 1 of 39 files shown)


src/Http/JsonLdResponse.php
```diff
@@ -30,7 +31,7 @@ public static function fromPsr7Response(ResponseInterface $response): self
     public function withData($data)
     {
         $body = $this->response->getBody();
-        $body->write(json_encode($data, self::JSON_OPTIONS));
+        $body->write(Json::encodeWithOptions($data, self::JSON_OPTIONS));
```
bertramakers, line 34: This is one of the most important usages because it determines how the JSON response of searches is encoded. It's interesting then that exactly this encode uses extra options which are not standardized in the `Json` class. Maybe we should add it as a `encodeForHttpResponse()` or something like that so it's clear it should be used in that context, and it will always use the right options for http responses then? Also applicable to udb3-silex I think, where we encode JSON for responses, but also for serializing things in our database or queuing messages for example. So it might be good to differentiate between context, like `encodeForHttpResponse()` and `encodeForSerialization()` for example?

bertramakers (approved):

src/Http/JsonLdResponse.php
LucWollants: For now I changed the specific case to `encodeForHttpResponse`, if we get more flavours better naming is indeed needed.

bertramakers: Yes in this app that is sufficient. I meant that in udb3-silex it might be good to add different flavors when we replace all usages of `json_encode()` there. I think right now we always use `Json::encode()` which just uses `THROW_ON_ERROR`.

PR #203, opened 2022-01-26
III-4424 Implement searching on regions for Organizers

LucWollants:
### Added
- Added `withRegionFilter` implementation inside `ElasticSearchOrganizerQueryBuilder`
- Added `withRegionFilter` call inside `OrganizerSearchController`
- Added `regions` as supported parameter
- Added `region` facet support

---


bertramakers (approved):
Nice work 👌

PR #207, opened 2022-04-21
III-4457 Make attendance mode searchable on events.

LucWollants:
### Added
- Added `attendanceMode` as an exact match field
- Added transformer for `attendanceMode` with default value of `offline`
- Added `attendanceMode` as a multiple values query parameter

---

Ticket: https://jira.uitdatabank.be/browse/III-4457


jdrieghe (approved):
Thanks for the clear incremental commits, this was a joy to review 🎉

PR #210, opened 2022-06-22
III-4788 Change error responses to return correct API problems

bertramakers:
### Changed
 
- Made sure non-existing routes return a 404 Not Found with API problem type `https://api.publiq.be/probs/url/not-found`
- Made sure a 405 Method Not Allowed uses API problem type `https://api.publiq.be/probs/method/not-allowed`
- Made sure an unknown query parameter (like `?foo=bar`) returns a 404 Not Found instead of a 400
- Made sure an unknown query parameter value (like `?addressCountry=bar`) returns a 404 Not Found instead of a 400
- Using an invalid URL as `website` parameter value now returns 0 results instead of a 400
 
---

Ticket: https://jira.uitdatabank.be/browse/III-4788
(See ticket for more examples)

### DIFF
--- a/src/ElasticSearch/Organizer/ElasticSearchOrganizerQueryBuilder.php
+++ b/src/ElasticSearch/Organizer/ElasticSearchOrganizerQueryBuilder.php
@@ -48,9 +48,20 @@ final class ElasticSearchOrganizerQueryBuilder extends AbstractElasticSearchQuer
         return $this->withMatchPhraseQuery('name.nl.autocomplete', $input);
     }
 
-    public function withWebsiteFilter(Url $url): ElasticSearchOrganizerQueryBuilder
+    public function withWebsiteFilter(string $url): ElasticSearchOrganizerQueryBuilder
     {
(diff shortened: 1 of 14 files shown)


src/ElasticSearch/Organizer/ElasticSearchOrganizerQueryBuilder.php
```diff
@@ -48,9 +48,20 @@ public function withAutoCompleteFilter(string $input): ElasticSearchOrganizerQue
         return $this->withMatchPhraseQuery('name.nl.autocomplete', $input);
     }
 
-    public function withWebsiteFilter(Url $url): ElasticSearchOrganizerQueryBuilder
-    {
-        return $this->withMatchQuery('url', $url->getNormalizedUrl());
+    public function withWebsiteFilter(string $url): ElasticSearchOrganizerQueryBuilder
```
LucWollants, line 51: Would it make sense to add a simple unit test that passes an invalid URL? Just to make sure the exception gets caught.

LucWollants (approved):
Very clean code and easy to review because of the granular commits.

PR #213, opened 2022-06-27
III-4806 Update `guzzlehttp/guzzle` to `v7.4`

bertramakers:
### Changed
 
- Updated `guzzlehttp/guzzle` to `v7.4`
- Updated `guzzlehttp/psr7` to `v2.4`
- Replaced `sentry/sdk` with latest version of `sentry/sentry`
 
---

Ticket: https://jira.uitdatabank.be/browse/III-4806


LucWollants (approved):
Looks good!

But now that I see the Sentry update I wonder if we should fake an error and see if it still arrives in Sentry?

PR #216, opened 2022-08-01
III-4822 Add consume command for related messages queue

bertramakers:
### Added
 
- Added consume command for "related" queue
 
---

Ticket: https://jira.uitdatabank.be/browse/III-4822

### DIFF
--- a/app/AmqpProvider.php
+++ b/app/AmqpProvider.php
@@ -17,7 +17,7 @@ final class AmqpProvider extends BaseServiceProvider
 {
     public function provides(string $alias): bool
     {
-        foreach ($this->getConsumers() as $consumerId => $consumerConfig) {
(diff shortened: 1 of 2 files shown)


bertramakers:
@willaerk the name for the new command will be `consume-udb3-related`, so for example `./bin/app.php consume-udb3-related` if running it from the root folder of the project

LucWollants (approved):

bertramakers:
@LucWollants @willaerk @JonasVHG I've refactored the code for the AMQP consumers a bit so they are now bootstrapped just based on the config. So we won't need to make PHP changes next time we want to add a consumer, just the config. The command names should remain the same so no supervisor code should break.

LucWollants (approved):

JonasVHG (approved):

PR #217, opened 2022-08-05
III-4834 Load default query from simple array config

LucWollants:
### Changed
- Load default queries from a simple config file containing an associative array. If the config does not contain an entry for the given API key a fallback is done to check UiTID
 
---

Ticket: https://jira.uitdatabank.be/browse/III-4834

### DIFF
--- a/src/Http/DefaultQuery/DefaultQueryConfigRepository.php
+++ b/src/Http/DefaultQuery/DefaultQueryConfigRepository.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search\Http\DefaultQuery;
+
+final class DefaultQueryConfigRepository implements DefaultQueryRepository
+{
+    private array $defaultQueryConfig;
+
+    public function __construct(array $defaultQueryConfig)
+    {
+        $this->defaultQueryConfig = $defaultQueryConfig;
+    }
+
+    public function getByApiKey(string $apiKey): ?string
+    {
+        return $this->defaultQueryConfig[$apiKey] ?? null;
+    }
+}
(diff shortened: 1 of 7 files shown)


JonasVHG (approved):

src/Http/DefaultQuery/DefaultQueryConfigRepository.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search\Http\DefaultQuery;
+
+final class DefaultQueryConfigRepository implements DefaultQueryRepository
```
bertramakers, line 7: Nit: Maybe `Config` is a bit vague here because this class does not load the actual config file, it just gets an array injected. Maybe `DefaultQueryArrayRepository` is clearer in this case? I usually use `InMemory...Repository` for repositories that just wrap an array, that's also an option.

bertramakers (approved):

src/Http/DefaultQuery/DefaultQueryConfigRepository.php
LucWollants: Indeed `Config` is too specific, I like the `InMemory` prefix.

bertramakers:
@LucWollants can you also add the `default_queries.php` to the `.gitignore`? Otherwise it will pop up as an untracked file when the initscripts put it in our local setup.

LucWollants:
> @LucWollants can you also add the `default_queries.php` to the `.gitignore`? Otherwise it will pop up as an untracked file when the initscripts put it in our local setup.

Good point, I added it in this PR https://github.com/cultuurnet/udb3-search-service/pull/218

PR #220, opened 2022-09-30
Build configuration for new Jenkins for UiTdatabank search service

willaerk:
### Added
 
- Jenkinsfile for job configuration
- Rake tasks for building artifacts

---

Ticket: https://jira.uitdatabank.be/browse/III-4993

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@
 /.php_cs.cache
 /.php-cs-fixer.cache
 /.default_queries.php
+/pkg
+/.bundle
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,7 @@
+source 'https://rubygems.org'
+
+gem 'rake', '~> 13.0'
+
+group :package do
+  gem 'fpm', '= 1.14.1'
+end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,0 +1,42 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    arr-pm (0.0.11)
+      cabin (> 0)
+    backports (3.23.0)
+    cabin (0.9.0)
+    clamp (1.0.1)
+    dotenv (2.7.6)
+    fpm (1.14.1)
+      arr-pm (~> 0.0.11)
+      backports (>= 2.6.2)
+      cabin (>= 0.6.0)
+      clamp (~> 1.0.0)
+      git (>= 1.3.0, < 2.0)
+      json (>= 1.7.7, < 3.0)
+      pleaserun (~> 0.0.29)
+      rexml
+      stud
+    git (1.11.0)
+      rchardet (~> 1.8)
+    insist (1.0.0)
+    json (2.6.1)
+    mustache (0.99.8)
+    pleaserun (0.0.32)
+      cabin (> 0)
+      clamp
+      dotenv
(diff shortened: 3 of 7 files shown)


paulherbosch (approved):

LucWollants:
@willaerk and @paulherbosch I wonder if it would be useful to put the Jenkins build files inside a hidden `.jenkins` folder as we do for `.github`.
That keeps the root of the project "cleaner" and focused on PHP code.

willaerk:
@LucWollants I'm not against that, but:
* that is the default location as noted in the official documentation:
> "In the Script Path field, specify the location (and name) of your Jenkinsfile. This location is the one that Jenkins checks out/clones the repository containing your Jenkinsfile, which should match that of the repository’s file structure. The default value of this field assumes that your Jenkinsfile is named "Jenkinsfile" and is located at the root of the repository."
* we have used this location everywhere possible up to now. If we change it, I would rather do it for all repositories/projects. And we then have to be explicit about the custom location in all pipelines. That would entail a lot of work.

LucWollants:
> That would entail a lot of work.
Is that work mainly situated in changing the repositories or on Jenkins? If it is mainly related to repositories I can help.

bertramakers:
I agree with @LucWollants that having these files mixed in with the PHP-related files feels a bit weird. I don't have 1 specific objective reason for it so that's why I didn't mention it before, but it does feel chaotic overall.

willaerk:
> is that work mainly situated in changing the repositories or on Jenkins?

It's both. We would need to move all Jenkinsfiles to the new locations, and then update all the pipelines to point to that new location. And then the first build after that will probably fail, as it creates somewhat of a chicken-egg problem: our pipelines are defined in the Jenkinsfile which are part of the repository we want to build an artifact from. But to instantiate the pipeline (or in other words start a build) we need a checkout of the code (and the contained Jenkinsfile).
Jenkins has a two step checkout flow (they call it lightweight checkout) that first fetches the Jenkinsfile and builds the pipeline steps in memory and then in a second phase checks out the rest of the code and proceeds with all the build steps.
Moving the Jenkinsfile will result in at least a failed first build, because it will search for the Jenkinsfile in the old location first.

After that, we would need to check (and probably change) all build steps for all projects, as they now assume to be running from the root directory of the repository. This is especially important for the creation and archiving of artifacts.

The Jenkinsfile is also just the automated progression of all steps of the project build, artifact build, artifact repository snapshot creation and publishing and the actual deployment. The separate steps are partly done in the rake tasks added and partly with API calls to different servers. If we change the directory of the Jenkinsfile we would also need to move the rake tasks, or at least make them aware of the changes. They would then need to reside in and reference the .jenkins directory, even when they are quite specifically written to be independent of Jenkins or any other build tool.

bertramakers:
If it's a lot of work to move them we definitely don't need to do that now IMO. There's no need to derail the work in progress (both on the new Jenkins and on Silex EOL / publiq platform) for a very small improvement. 

But I do have several other concerns with this approach as I shared with you on Slack @willaerk, and this problem of it being a lot of work to go over every repository to make a small change to the setup kind of reinforces them.

I propose we leave it like it is for now and continue this approach, but I would also plan a meeting to discuss how we ended up here and how we can avoid these discussions in the future. Sound good to everyone?

PR #239, opened 2023-02-22
Docker search service

JonasVHG:
### Added
 
- `Docker.md` with setup instructions.
- `Makefile` to easily run command on Docker container
- `Dockerfile` to create container
- `docker-compose` to run container
 
### Changed
 
- `.gitignore` to work with docker setup (see docker.md for reason)

---

Ticket: https://jira.uitdatabank.be/browse/...


docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+      uitdatabank:
+        aliases:
+            - elasticsearch.uitdatabank.dev
+
+networks:
+  uitdatabank:
+    driver: bridge
+
+volumes:
+  udb3-mysql:
```
LucWollants, line 43: There is no need for the `mysql` volume here, but maybe we need one for `es`?

docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+        php bin/app.php consume-udb3-related &
+        php -S 0.0.0.0:9000 -t web
+    ports:
+      - "9000:9000"
+    networks:
+      uitdatabank:
+        aliases:
+          - search.uitdatabank.dev
+
+  elasticsearch:
```
LucWollants, line 26: Maybe this also needs a container name? Something like `container_name: es.uitdatabank`

docker.md
```diff
@@ -0,0 +1,76 @@
+# UDB3 Search Service with Docker
+
+## Prerequisite
+- Install Docker Desktop
+
+## Configure
+
+### config.yml file
+
+Copy the latest `config.yml` from https://github.com/cultuurnet/udb3-vagrant/tree/main/config/udb3-search-service to the root
```
LucWollants, line 10: Do we also need to copy the `facet` config files?

docker.md
```diff
@@ -0,0 +1,76 @@
...
+- amqp.port
+- amqp.user
+- amqp.password
+- amqp.vhost
+
+To read messages from [udb3-backend](https://github.com/cultuurnet/udb3-backend):
+- Create an exchange `udb3.x.domain-events` in your RabbitMQ provider
+- Make the url of the docker-container of [udb3-backend](https://github.com/cultuurnet/udb3-backend) available
+to the docker container of [udb3-search-service](https://github.com/cultuurnet/udb3-search-service)
+  (one possible way is using `host.docker.internal` as `url` in [udb3-backend](https://github.com/cultuurnet/udb3-backend)
```
LucWollants, line 40: I think this belongs to the `udb3-backend` readme and also requires a change to the acceptance tests configuration?

docker.md
```diff
@@ -0,0 +1,76 @@
...
+In your `config.yml` file, you have to change some of the hosts to work with Docker instead of Vagrant.
+
+You'll need to change the following lines to work with docker hosts:
+- elasticsearch.host: 
+    - `elasticsearch`
+- geoshapes_index.indexation.path:
+    - `/geojson-data/output/elasticsearch`
+
+### ElasticSearch
+To setup ElasticSearch you'll have to clone [geojson-data](https://github.com/cultuurnet/geojson-data)
```
LucWollants, line 21: Where should this be cloned, inside the root or a sub folder?

docker.md
```diff
@@ -0,0 +1,76 @@
...
+## Prerequisite
+- Install Docker Desktop
+
+## Configure
+
+### config.yml file
+
+Copy the latest `config.yml` from https://github.com/cultuurnet/udb3-vagrant/tree/main/config/udb3-search-service to the root
+
+In your `config.yml` file, you have to change some of the hosts to work with Docker instead of Vagrant.
```
LucWollants, line 12: Some of the hosts? It seems that only one host needs to be changed?

docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+      - .:/var/www/html:cached
+    working_dir: /var/www/html
+    command:
+      - /bin/bash
+      - -c
+      - |
+        php bin/app.php consume-udb3-api &
+        php bin/app.php consume-udb3-cli &
+        php bin/app.php consume-udb3-related &
+        php -S 0.0.0.0:9000 -t web
```
LucWollants, line 18: Does `udb3-backend` uses port `9000` for the search requests?

docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+      uitdatabank:
+        aliases:
+          - search.uitdatabank.dev
+
+  elasticsearch:
+    image: public.ecr.aws/y2o7k5u2/udb3/elasticsearch:5.3.3
+    restart: always
+    ports:
+      - '9200:9200'
+    expose:
```
LucWollants, line 31: I don't think `expose` is needed, it is overruled by `ports`.

See for difference between `ports` and `expose` https://stackoverflow.com/questions/40801772/what-is-the-difference-between-ports-and-expose-in-docker-compose

docker-compose.yml
JonasVHG: Indeed, could be handy

.gitignore
```diff
@@ -1,6 +1,7 @@
 /vendor
 /build
 /log
+/geojson-data
```
JonasVHG, line 4: I've tried to work with a folder for `geojson-data` out of `udb3-search-service` but this gives errors. Still it would be worth to get it to work.

bertramakers (approved):

bertramakers:
I accidentally approved this PR already because I got a notification about the latest commit and thought that was a (small) PR on its own.

.gitignore
LucWollants: It works now with `geojson-data` cloned next to `udb3-search-service`

docker-compose.yml
LucWollants: Port 9000 is required to serve the SAPI3 proxy next to port 8000 for the udb3-backend API

docker.md
LucWollants: Added extra info about configs.

docker.md
LucWollants: Moved to `udb3-backend`

docker.md
LucWollants: Updated the readme

docker.md
LucWollants: Updated readme

LucWollants (approved):

PR #241, opened 2023-02-24
III-5326 search contributors param

JonasVHG:
### Added
 
- `ContributorsTransformer` , `Offer\ContributorsRequestParser`, `Organizer\ContributorsRequestParser`
- Added `FilterParam` & `Index` for `contributors` on `events`, `places` & `organizers`

### Changed

- added `contributors` to `removeInternalProperties()` in `JsonLdPolyfillJsonTransformer` 
 
### TODO: in following PR
- Refactor `GuzzleJsonDocumentFetcher` to store `Auth0Token` to avoid reaching our quota.
---

### DIFF
--- a/src/JsonDocument/GuzzleJsonDocumentFetcher.php
+++ b/src/JsonDocument/GuzzleJsonDocumentFetcher.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\Search\JsonDocument;
 
+use CultuurNet\UDB3\Search\Http\Authentication\Auth0Client;
 use CultuurNet\UDB3\Search\ReadModel\JsonDocument;
 use GuzzleHttp\ClientInterface;
 use Psr\Log\LoggerInterface;
@@ -16,11 +17,14 @@ final class GuzzleJsonDocumentFetcher implements JsonDocumentFetcher
 
     private LoggerInterface $logger;
 
-    public function __construct(ClientInterface $httpClient, LoggerInterface $logger)
+    private Auth0Client $auth0Client;
+
+    public function __construct(ClientInterface $httpClient, LoggerInterface $logger, Auth0Client $auth0Client)
     {
         $this->httpClient = $httpClient;
         $this->includeMetadata = false;
         $this->logger = $logger;
+        $this->auth0Client = $auth0Client;
     }
 
     public function withIncludeMetadata(): JsonDocumentFetcher
@@ -35,7 +39,7 @@ final class GuzzleJsonDocumentFetcher implements JsonDocumentFetcher
         $response = $this->httpClient->request(
             'GET',
             $documentIri,
-            $this->getQuery($this->includeMetadata)
+            array_merge($this->getQuery($this->includeMetadata), $this->getHeader())
         );
 
         if ($response->getStatusCode() !== 200) {
@@ -64,6 +68,20 @@ final class GuzzleJsonDocumentFetcher implements JsonDocumentFetcher
             'query' => [
                 'includeMetadata' => true,
                 'embedUitpasPrices' => true,
+            ],
+        ];
+    }
+
+    private function getHeader(): array
+    {
+        $token = $this->auth0Client->getToken();
+        if ($token === null) {
+            return [];
+        }
+
+        return [
+            'headers' => [
+                'Authorization' => 'Bearer ' . $token->getToken(),
             ],
         ];
     }
--- a/src/JsonDocument/GuzzleJsonDocumentFetcher.php
+++ b/src/JsonDocument/GuzzleJsonDocumentFetcher.php
@@ -10,20 +10,11 @@ use Psr\Log\LoggerInterface;
 
 final class GuzzleJsonDocumentFetcher implements JsonDocumentFetcher
 {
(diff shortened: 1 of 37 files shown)


bertramakers (commented):
@JonasVHG all the changes so far look great, I just can't test them because the calls to io to fetch the JSON are not authenticated yet so the `contributors` are never actually included. Can you implement that as well?

Also, you'll need to update the version of the `UDB3_CORE` schema in `SchemaVersions` to make it possible to apply the new mapping (as documented in https://github.com/cultuurnet/udb3-search-service#indexing-a-new-field)

LucWollants:
I also think we are missing several unit tests for the transformers and query builders, for example:
- ElasticSearchOfferQueryBuilderTest
- ElasticSearchOrganizerQueryBuilderTest
- EventTransformerTest
- OrganizerTransformerTest

src/JsonDocument/GuzzleJsonDocumentFetcher.php
```diff
@@ -76,4 +71,18 @@ private function getQuery(bool $includeMetadata): array
             ],
         ];
     }
+
+    private function getHeader(): array
+    {
+        $token = $this->auth0Client->getToken();
```
bertramakers, line 77: I completely forgot that we already had an Auth0 client class in this repository, good idea to re-use that. However, this method will always return a new token. Instead, we should re-use the same token until it's expired. We can check that it's expired when Entry API returns a 401 response instead of a 200 response.

We need to do that because we have a quota on machine-to-machine tokens in Auth0, so every client that uses those tokens should cache them.

Usually, we should cache the token in e.g. Redis to do this correctly. However, because this code that fetches documents from Entry API only runs in the AMQP consumer that runs for a long continuous time, we can store it in-memory instead. So I would suggest to store the token in a property in this class, and always use that until Entry API returns a 401 and then get a new token and re-try the request.

Note that the Guzzle client will by default throw an HttpException when a 4xx or 5xx response is returned. You can catch that, or disable it with the `http_errors` option. https://docs.guzzlephp.org/en/stable/request-options.html#http-errors

JonasVHG: As discussed on Slack, I will do this is a seperate PR.

tests/ElasticSearch/JsonDocument/data/event/indexed-with-contributors.json
```diff
@@ -0,0 +1,79 @@
...
+        "@type": "Place",
+        "id": "179c89c5-dba4-417b-ae96-62e7a12c2405",
+        "mainId": "179c89c5-dba4-417b-ae96-62e7a12c2405",
+        "name": {
+            "nl": "Hungaria"
+        }
+    },
+    "created": "2017-04-22T13:33:37+02:00",
+    "creator": "Jane Doe",
+    "originalEncodedJsonLd": "{\"@id\":\"http:\/\/udb-silex.dev\/event\/23017cb7-e515-47b4-87c4-780735acc942\",\"mainLanguage\":\"nl\",\"languages\":[\"nl\"],\"completedLanguages\":[\"nl\"],\"name\":{\"nl\":\"Punkfest\"},\"calendarType\":\"single\",\"startDate\":\"2017-04-22T10:00:00+00:00\",\"endDate\":\"2017-04-22T18:00:00+00:00\",\"availableTo\":\"2017-04-22T18:00:00+00:00\",\"location\":{\"@id\":\"http:\/\/udb-silex.dev\/place\/179c89c5-dba4-417b-ae96-62e7a12c2405\",\"name\":{\"nl\":\"Hungaria\"},\"address\":{\"nl\":{\"addressCountry\":\"BE\",\"addressLocality\":\"Leuven\",\"postalCode\":\"3000\",\"streetAddress\":\"Vaartkom 35\"}}},\"workflowStatus\":\"DRAFT\",\"created\":\"2017-04-22T13:33:37+02:00\",\"creator\":\"Jane Doe\",\"contributors\":[\"amy@example.com\",\"john@example.com\"]}",
```
LucWollants, line 69: I think `contributors` should not be part of the originalEncodedJsonLd just like `metadata`?

JonasVHG: But `original-with-contributors.json` didn't have `metadata` so I does not show up in `indexed-with-contributors.json`.
If you look at https://github.com/cultuurnet/udb3-search-service/blob/15b68ea68b0b54331061fc7d9f900401a3ed65cf/tests/ElasticSearch/JsonDocument/data/event/indexed-with-metadata.json#L69 I assume `metadata` should be in `originalEncodedJsonLd`.

LucWollants: I would expect that this line removes it?
https://github.com/cultuurnet/udb3-search-service/pull/241/commits/9545f8858887ebe9e620e44e911191b5d6319636

LucWollants: After looking into the code. The original JSON gets stored inside the index as is, when serving it extra code will remove it. So the test is correct.

LucWollants (approved):

PR #242, opened 2023-03-01
III-5326 refactor guzzle json document fetcher

JonasVHG:
### Changed
 
- `GuzzleJsonDocumentFetcher` store `Auth0Token` and only renew it after a `401`-Request
- `JsonDocumentFetcherTest` added test to check if tokens are refreshed
- `JsonDocumentFetcherTest` added test to check that tokens are not infinitely refreshed
 
---

### DIFF
--- a/src/JsonDocument/GuzzleJsonDocumentFetcher.php
+++ b/src/JsonDocument/GuzzleJsonDocumentFetcher.php
@@ -41,12 +41,8 @@ final class GuzzleJsonDocumentFetcher implements JsonDocumentFetcher
 
     public function fetch(string $documentId, string $documentIri): ?JsonDocument
     {
-        $attempt = 0;
-        $maximumAttempts = 2;
-
         $response = $this->getResponse($documentIri);
-        while ($response->getStatusCode() === 401 && $attempt < $maximumAttempts) {
-            $attempt++;
+        if ($response->getStatusCode() === 401) {
             $this->refreshToken();
             $response = $this->getResponse($documentIri);
         }
--- a/src/JsonDocument/GuzzleJsonDocumentFetcher.php
+++ b/src/JsonDocument/GuzzleJsonDocumentFetcher.php
@@ -5,8 +5,10 @@ declare(strict_types=1);
 namespace CultuurNet\UDB3\Search\JsonDocument;
 
 use CultuurNet\UDB3\Search\Http\Authentication\Auth0Client;
+use CultuurNet\UDB3\Search\Http\Authentication\Auth0Token;
 use CultuurNet\UDB3\Search\ReadModel\JsonDocument;
 use GuzzleHttp\ClientInterface;
+use Psr\Http\Message\ResponseInterface;
 use Psr\Log\LoggerInterface;
 
 final class GuzzleJsonDocumentFetcher implements JsonDocumentFetcher
@@ -19,12 +21,15 @@ final class GuzzleJsonDocumentFetcher implements JsonDocumentFetcher
 
     private Auth0Client $auth0Client;
 
+    private ?Auth0Token $auth0Token;
+
     public function __construct(ClientInterface $httpClient, LoggerInterface $logger, Auth0Client $auth0Client)
     {
         $this->httpClient = $httpClient;
         $this->includeMetadata = false;
         $this->logger = $logger;
         $this->auth0Client = $auth0Client;
+        $this->auth0Token = $this->auth0Client->getToken();
     }
 
     public function withIncludeMetadata(): JsonDocumentFetcher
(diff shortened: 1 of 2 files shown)


src/JsonDocument/GuzzleJsonDocumentFetcher.php
```diff
@@ -36,11 +41,15 @@ public function withIncludeMetadata(): JsonDocumentFetcher
...
-        $response = $this->httpClient->request(
-            'GET',
-            $documentIri,
-            array_merge($this->getQuery($this->includeMetadata), $this->getHeader())
-        );
+        $attempt = 0;
+        $maximumAttempts = 2;
+
+        $response = $this->getResponse($documentIri);
+        while ($response->getStatusCode() === 401 && $attempt < $maximumAttempts) {
```
LucWollants, line 48: What is the reasoning behind doing 2 attempts?

LucWollants: Why not just make the request with the existing token, in the case of 401 refresh the token and make the request again. When that request with the refreshed tokens fails log it and bailout.

JonasVHG: I was being extra prudent, but indeed if the first `refresth()` fails, the second `refresh()` should normally also always fail.

tests/JsonDocument/JsonDocumentFetcherTest.php
```diff
@@ -251,7 +262,167 @@ public function it_can_authorize_requests(): void
...
+        $authorizedJsonDocumentFetcher->fetch(
+            $documentId,
+            $documentUrl
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function it_will_not_refresh_tokens_infinitely(): void
```
LucWollants, line 352: Maybe a better name is something like `it_fails_when_refreshed_token_is_invalid` ?

tests/JsonDocument/JsonDocumentFetcherTest.php
```diff
@@ -251,7 +262,167 @@ public function it_can_authorize_requests(): void
...
+                    ],
+                ]
+            )
+            ->willReturnOnConsecutiveCalls(
+                new Response(
+                    200,
+                    [],
+                    Json::encode([
+                        'access_token' => self::DUMMY_TOKEN,
+                        'expires_in' => 1,
```
LucWollants, line 299: Not related to this pull request, but I wonder why `expires_in` was added previously? Could we just remove it because everything related to HTTP calls seems to be mocked?

JonasVHG: It is used in `Auth0TokenProvider` to fetch a new token if it will expire within 5 minutes. https://github.com/cultuurnet/udb3-search-service/blob/15b68ea68b0b54331061fc7d9f900401a3ed65cf/src/Http/Authentication/Auth0TokenProvider.php#L31
So it will take some refactoring to remove it, I guess.

LucWollants: But I mean in the context of the tests, because it is mocked that a `401` will be returned.

tests/JsonDocument/JsonDocumentFetcherTest.php
```diff
@@ -251,7 +262,167 @@ public function it_can_authorize_requests(): void
...
+                    200,
+                    [],
+                    Json::encode([
+                        'access_token' => self::DUMMY_TOKEN,
+                        'expires_in' => 10000,
+                    ])
+                ),
+            );
+
+        $authorizedJsonDocumentFetcher = (new GuzzleJsonDocumentFetcher(
```
LucWollants, line 312: Why is a new document fetcher created?

JonasVHG: Because the auth0httpClient is being mocked in `setup()` and will return null on gettoken.
An alternative could be to move the `$this->auth0httpClient->expects($this->once())
            ->method('post')`
to the setup, but then it might be difficult to count the calls it makes.

LucWollants: I think it is related to the extra call inside the constructor `$this->auth0Client->getToken()`

LucWollants: I think we should move the call from the constructor the `fetch` method. It is even tricky in a real scenario to call it inside the constructor.

tests/JsonDocument/JsonDocumentFetcherTest.php
LucWollants: I now noticed that the Token is created from the JSON body and it requires an `expires_in` property. So it it required for this line of code https://github.com/cultuurnet/udb3-search-service/blob/main/src/Http/Authentication/Auth0Client.php#L76

LucWollants (approved):

PR #243, opened 2023-03-01
III-5326 Refactor Auth0client

JonasVHG:
### Changed
 
-  Added optional parameter `$audience` to `Auth0Client` to distinguish between management & non-management client.
- Add `$audience` to `Auth0Client` in `GuzzleJsonDocumentFetcher`

---


LucWollants (changes requested):
I think you could remove the new constructor argument `isManagementClient` and inject the correct domains during instantiation. That requires fewer code changes.

Now that I read up on Slack about the proposed solution, should the code switch between `https://publiq-acc.eu.auth0.com/api/v2` for Auth0 Management and `https://api.publiq.be/` for API usage

LucWollants (approved):

PR #245, opened 2023-03-07
III-5414 Add hostname property to avoid random node names

LucWollants:
### Added
 - Added hostname property to avoid random node names (in combination with new ES image with correct owner of the data folder)
 
---

Ticket: https://jira.uitdatabank.be/browse/III-5414


JonasVHG (approved):
persistence works👍

PR #246, opened 2023-03-07
Use custom RabbitMQP setup

LucWollants:
### Changed
- Used custom RabbitMQ image with preconfigured exchange and queues equal to Vagrant setup

### DIFF
--- a/docker.md
+++ b/docker.md
@@ -17,6 +17,10 @@ You'll need to change the following lines to work with docker hosts:
 elasticsearch.host: 
   elasticsearch
 ```
+```
+amqp.host: 
+  host.docker.internal
+```
 
 Copy all the latest facet files from https://github.com/cultuurnet/udb3-vagrant/tree/main/config/udb3-search-service
 
--- a/docker.md
+++ b/docker.md
@@ -30,19 +30,7 @@ Copy the `public-auth0.pem` from https://github.com/cultuurnet/udb3-vagrant/blob
 
 ### RabbitMQ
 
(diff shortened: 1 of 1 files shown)


docker.md
```diff
@@ -30,19 +30,7 @@ Copy the `public-auth0.pem` from https://github.com/cultuurnet/udb3-vagrant/blob
...
-  port: 5672
-  vhost: /
-  user: guest
-  password: guest
-```
-
-To read messages from [udb3-backend](https://github.com/cultuurnet/udb3-backend):
-- Create an exchange `udb3.x.domain-events` in your RabbitMQ provider
-- Communication between Docker containers is handled through `host.docker.internal`
+Login to the management console on http://host.docker.internal:15672/ with username `vagrant` and password `vagrant`
```
JonasVHG, line 33:
```suggestion
Login to the management console on http://host.docker.internal:15672/ with username `guest` and password `guest`
```

JonasVHG (changes requested):
There seems to be a small typo in the login data of the management console.

docker.md
LucWollants: No, the [custom RabbitMQ image](https://github.com/cultuurnet/udb3-backend/pull/1288/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R50) used in `udb3-backend` has everything set up like the Vagrant box. So maybe it's a bit confusing but vagrant is used for username, password, and host 😉

JonasVHG: My mistake 🫣, I only deleted my search-containers not my rabbitMq container.

JonasVHG (approved):

PR #247, opened 2023-03-07
III-4590 Organizers id filter

JonasVHG:
### Added
 
- Added `id` to the mapping for `Organizers`
- Added `id` to `OrganizerSupportedParameters`
- `IdRequestParser` for `Organizers`
 
### Changed
 
- Updated the `UDB3_CORE` version number
- `id` is now valid parameter for `Organizers`
- `OrganizerSearchServiceProvider`: injection of `IdRequestParser`
 
---

Ticket: https://jira.uitdatabank.be/browse/III-4590

### DIFF
--- a/src/Http/Organizer/RequestParser/IdRequestParser.php
+++ b/src/Http/Organizer/RequestParser/IdRequestParser.php
@@ -14,10 +14,10 @@ final class IdRequestParser implements OrganizerRequestParser
         OrganizerQueryBuilderInterface $organizerQueryBuilder
     ): OrganizerQueryBuilderInterface {
         $parameterBagReader = $request->getQueryParameterBag();
-        $cdbid = $parameterBagReader->getStringFromParameter('id');
+        $organizerId = $parameterBagReader->getStringFromParameter('id');
 
-        if (!is_null($cdbid)) {
-            $organizerQueryBuilder = $organizerQueryBuilder->withIdFilter($cdbid);
+        if (!is_null($organizerId)) {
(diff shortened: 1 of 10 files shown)


LucWollants (approved):
Looks good! I only think a test is missing in `OrganizerSearchControllerTest`

bertramakers (commented):
Aside from Luc's suggestion I only have a very small nit to add. I'd personally try to avoid adding more usages of the term "cdbid". I know there is already an existing "Cdbid" class in this codebase (which I probably even added myself years ago), but the abbreviation is kind of outdated ("cultuurdatabank id") and almost never used anymore in our team(s) or for example the APIs. We usually just use "id" or more specifically "eventId", "placeId", etc. I usually only see it used in other teams that have gotten so used to the term over the years, especially among less technical people, or older systems where it was never changed. But it's a very minor implementation detail, I'm just sharing it as a general thought.

Edit to add: Given that the `Cdbid` class just extends `StringLiteral` without adding any logic, for me it would also be okay to just use a scalar `string` instead since we have typing for that these days. (I'd certainly not _add_ validation logic for the ids, since they'd ideally be considered opaque strings without a specific format to avoid issues if they'd ever change, which is not totally unlikely for example if we ever don't just use UUID v4 but also v5 or for example the NIL UUID that we already use)

src/Http/Organizer/RequestParser/IdRequestParser.php
```diff
@@ -18,7 +17,7 @@ public function parse(
         $cdbid = $parameterBagReader->getStringFromParameter('id');
```
LucWollants: Here is still a local variable called `$cdbid`

PR #252, opened 2023-07-18
Added features.yml in the documentation docker.md

grubolsch:
### Added
 
- Added features.yml mention in the docker.md file

### DIFF
--- a/docker.md
+++ b/docker.md
@@ -6,7 +6,9 @@
 
 ## Configure
 
-### config.yml file
+### Configuration files.
+
+You will have to copy a couple of configuration yml files, place them in the root of the udb3-search-service project.
 
 Copy the latest `config.yml` from https://github.com/cultuurnet/udb3-vagrant/tree/main/config/udb3-search-service to the root
 
@@ -27,6 +29,8 @@ Copy all the latest facet files from https://github.com/cultuurnet/udb3-vagrant/
 - facet_mapping_facilities.yml
 - facet_mapping_themes.yml
 - facet_mapping_types.yml
+
+Copy the feature file from https://github.com/cultuurnet/udb3-vagrant/tree/main/config/udb3-search-service/features.yml
 
 Copy all the regions from https://github.com/cultuurnet/geojson-data/blob/main/output/facet_mapping_regions.yml
 


LucWollants:
I think we should also place the config files inside the docker folder inside appconfig like we did for `udb3-backend` https://github.com/cultuurnet/appconfig/tree/main/files/udb3/docker/udb3-backend

The `udb3-backend` project has a script to copy the config https://github.com/cultuurnet/udb3-backend/blob/master/docker/config.sh. I think it would be nice to also do it for this service.

grubolsch:
I added a script based on Luc his comment to automate a lot of the steps in docker.md

bin/config.sh
```diff
@@ -0,0 +1,25 @@
...
+DIR="../udb3-vagrant/config/"
+if [ -d "$DIR" ]; then
+  cp "$DIR"/udb3-search-service/config.yml .
+  cp "$DIR"/udb3-search-service/facet_mapping_facilities.yml .
+  cp "$DIR"/udb3-search-service/facet_mapping_themes.yml .
+  cp "$DIR"/udb3-search-service/facet_mapping_types.yml .
+  cp "$DIR"/udb3-search-service/features.yml .
+  cp "$DIR"/keys/public-auth0.pem .
+
+  sed -i '' 's!http://elasticsearch.dev:9200!elasticsearch!g' config.yml
```
LucWollants, line 12: Instead of replacing it we could have dedicated docker config files like we have for `udb3-backend`

bin/config.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+
+DIR="../udb3-vagrant/config/"
```
LucWollants, line 3: This should point to the docker config files for udb3-search-service

grubolsch: Is this not a problem because this repo is public, but the appconfig repo is private? Making it not possible for outside developers to install the application?

JonasVHG: `udb3-vagrant` is also a private repo. We cannot make these public because they contain keys.

LucWollants: It is correct that external developers can have issues getting the correct config. But we never had a request for this, so when the need arises we could solve it with example configs.

grubolsch: Oké, I updated the script and added some new configs in the appconfig repo.
See the sister PR: https://github.com/cultuurnet/appconfig/pull/394

grubolsch:
Oké, I updated the script and added some new configs in the appconfig repo.
See the sister PR: https://github.com/cultuurnet/appconfig/pull/394

bin/config.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+DIR="../appconfig/files/udb3/docker"
+if [ -d "$DIR" ]; then
+  cp "$DIR"/udb3-search-service/config.yml .
```
LucWollants, line 5: Now that everything is inside the `docker` folder, the full content can be copied?

bin/config.sh
```diff
@@ -0,0 +1,22 @@
...
+  cp "$DIR"/udb3-search-service/facet_mapping_facilities.yml .
+  cp "$DIR"/udb3-search-service/facet_mapping_themes.yml .
+  cp "$DIR"/udb3-search-service/facet_mapping_types.yml .
+  cp "$DIR"/udb3-search-service/features.yml .
+  cp "$DIR"/udb3-backend/public-auth0.pem .
+  else
+  echo "Error: missing udb3-vagrant see docker.md prerequisites to fix this."
```
LucWollants, line 12: Missing `docker` instead of missing `ud3-vagrant`?

docker.md
```diff
@@ -1,36 +1,15 @@
...
-Copy all the latest facet files from https://github.com/cultuurnet/udb3-vagrant/tree/main/config/udb3-search-service
-
-- facet_mapping_facilities.yml
-- facet_mapping_themes.yml
-- facet_mapping_types.yml
-
-Copy all the regions from https://github.com/cultuurnet/geojson-data/blob/main/output/facet_mapping_regions.yml
-
-Copy the `public-auth0.pem` from https://github.com/cultuurnet/udb3-vagrant/blob/main/config/keys/public-auth0.pem
+Run the script `./bin/config.sh`, this will copy a couple of configuration yml files, placing them in the root of the udb3-search-service project, and modify the config.yml for local development.
```
LucWollants, line 12: I think the script is inside the `docker` folder inside `ubd3-backend`

LucWollants (changes requested):
Looks good! Only some minor tweaks to make it exactly the same as `udb3-backend`

docker.md
grubolsch: No, the config.sh is inside `./bin/config.sh`, this path is correct

LucWollants: Yes, here it is in the bin folder, but I wanted to point out that for `udb3-backend` it's inside another location. So it's better to use the docker location because it only makes sense in the docker context.

grubolsch: oké, I get it now. it has been changed.

LucWollants (approved):

PR #255, opened 2023-11-15
III-5875 Embed contributers to uri

grubolsch:
### Changed
Added embedContributors GET param to URI to fetch document for indexing

---

Ticket: https://jira.uitdatabank.be/browse/III-5875


grubolsch:
I am aware a unit test is broken, but I think the unit test is actually wrong in how the behavior should be.
The contributers should NOT be included in the originalEncodedJsonLd, because then everybody has access to it.

LucWollants:
> I am aware a unit test is broken, but I think the unit test is actually wrong in how the behavior should be. The contributers should NOT be included in the originalEncodedJsonLd, because then everybody has access to it.

@grubolsch I think your assumption is correct.

LucWollants:
@grubolsch but I also found this method `JsonLdPolyfillJsonTransformer::removeInternalProperties`

grubolsch:
All feedback has been processed.

LucWollants (approved):

PR #271, opened 2024-04-02
III-6128 upgrade phpunit and friends to 9.6

grubolsch:
### Changed
Upgraded phpunit to 9.6
Updated / removed some deprecations in the tests because of this upgrade.

---

Ticket: https://jira.uitdatabank.be/browse/III-6128

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "cakephp/chronos": "^1.1",
         "crell/api-problem": "^3.2",
         "cultuurnet/calendar-summary-v3": "^4.0",
-        "cultuurnet/culturefeed-php": "~1.12",
+        "cultuurnet/culturefeed-php": "^1.14",
         "filp/whoops": "^2.5",
         "guzzlehttp/guzzle": "^7.4",
         "guzzlehttp/psr7": "^2.4",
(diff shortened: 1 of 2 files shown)


tests/AMQP/EventBusForwardingConsumerTest.php
```diff
@@ -136,32 +142,26 @@ function ($domainEventStream) use ($expectedMetadata, $expectedPayload) {
...
-            ->expects($this->at(0))
-            ->method('info')
-            ->with(
-                'received message with content-type application/vnd.cultuurnet.udb3-events.dummy-event+json',
-                $context
-            );
-
+        $messageLog = [];
         $this->logger
-            ->expects($this->at(1))
```
LucWollants, line 151: This is an example of how it was fixed on `publiq-platform` by using a `match` statement and keeping the code more contained.

grubolsch: match() is PHP8, so I cannot do that if I want 7.4 compatibility.
My code also does a little bit more because it makes sure all messages have been called once - just like the old tests.

simon-debruijn (approved):

PR #277, opened 2024-04-03
Iii 6128/upgrade to phpstan lvl 4

grubolsch:
Changed
Changed phpstan lvl to lvl 4.
Improved the code so it could pass.

Ticket: https://jira.uitdatabank.be/browse/III-6128


src/ElasticSearch/JsonDocument/Properties/CalendarTransformer.php
```diff
@@ -334,7 +329,7 @@ private function polyFillJsonLdSubEventsFromOpeningHours(array $from): array
     /**
      * @param array $openingHours
      *   JSON-LD of the openingHours property of an event/place, as an associative array
-     * @return array<int,array<int,array<int,string>>>
+     * @return array<string, array<int<0, max>, array<string, mixed>>>
```
simon-debruijn, line 332: Cool that you can type positive numbers in phpstan

app/Console/InstallUDB3CoreCommand.php
```diff
@@ -15,31 +15,15 @@
...
-    /**
-     * @var string
-     */
-    private $latestIndexName;
-
-    /**
-     * @var string
-     */
-    private $writeAlias;
+    private string$latestIndexName;
```
LucWollants, line 18:
```suggestion
    private string $latestIndexName;
```

app/Console/InstallUDB3CoreCommand.php
```diff
@@ -88,7 +72,7 @@ protected function execute(InputInterface $input, OutputInterface $output): ?int
             return 0;
         }
 
-        if ($latestIndexExists && $force) {
+        if ($latestIndexExists) {
```
LucWollants, line 75: The `$force` parameter is also removed here.

src/ElasticSearch/IndexationStrategy/BulkIndexationStrategy.php
```diff
@@ -17,9 +17,6 @@ final class BulkIndexationStrategy implements IndexationStrategy
 
     private int $autoFlushThreshold;
 
-    /**
-     * @var JsonDocument[]
```
LucWollants, line 21: This can be useful information that seems to be removed now.

src/Http/Authentication/JsonWebToken.php
```diff
@@ -22,13 +22,16 @@ final class JsonWebToken
 
     public function __construct(string $jwt)
     {
-        $this->token = (new Parser(new JoseEncoder()))->parse($jwt);
+        // Dirty: This docblock is needed for phpstan - a more elegant solution is welcome.
+        /** @var UnencryptedToken $token */
+        $token = (new Parser(new JoseEncoder()))->parse($jwt);
```
LucWollants, line 27: I had to do the same change in udb3-backend, see https://github.com/cultuurnet/udb3-backend/blob/master/src/Http/Auth/Jwt/JsonWebToken.php#L45-L46

src/JsonDocument/GuzzleJsonDocumentFetcher.php
```diff
@@ -32,14 +32,16 @@ public function __construct(ClientInterface $httpClient, LoggerInterface $logger
         $this->auth0Client = $auth0Client;
     }
 
-    public function withIncludeMetadata(): JsonDocumentFetcher
+    /** @return static */
+    public function withIncludeMetadata()
```
LucWollants, line 36: I wonder why this was changed would prefer a return of `GuzzleJsonDocumentFetcher`

src/Http/Authentication/JsonWebToken.php
simon-debruijn: I found a nice alternative when I was reading phpstan docs yesterday

https://phpstan.org/writing-php-code/phpdocs-basics#inline-%40var

assertint the runtime class instance is of the correct type you assume

tests/AMQP/Dummies/DummyEventNotSerializable.php
```diff
@@ -6,23 +6,24 @@
...
-     * @param string $content
-     */
-    public function __construct($id, $content)
+    public function __construct(string $id, string $content)
     {
         $this->id = $id;
         $this->content = $content;
     }
+
+    // These getters are never used, just here so phpstan is happy
```
LucWollants, line 18: I removed them locally and PHPStan was still happy

src/Http/Authentication/JsonWebToken.php
simon-debruijn:
```php
<?php

class Test {
	public static function test(): void {}
}

function createTest(): mixed {
	return new Test();
}

$foo = createTest();
assert($foo instanceof Test);

$foo::test();
```

LucWollants: That is indeed a smart solution, so it becomes something like
```
    public function __construct(string $jwt)
    {
        $token = (new Parser(new JoseEncoder()))->parse($jwt);
        assert($token instanceof UnencryptedToken, 'Token should be an instance of UnencryptedToken');
        $this->token = $token;
    }
```

tests/Facet/FacetTreeTest.php
```diff
@@ -97,41 +96,21 @@ public function it_has_a_multi_level_list_of_facet_node_children(): void
...
     }
 
-
     private function assertChildrenEquals(array $expected, array $actual): void
     {
-        $this->assertEquals(count($expected), count($actual));
+        $this->assertSameSize($expected, $actual);
 
-        for ($i = 0; $i < count($expected); $i++) {
+        for ($i = 0, $iMax = count($expected); $i < $iMax; $i++) {
```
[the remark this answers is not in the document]
grubolsch: Yes, it is to prevent recalulation - it improves performance.
I move it to outside the loop.

app/Console/InstallGeoShapesCommand.php
```diff
@@ -88,7 +74,7 @@ protected function execute(InputInterface $input, OutputInterface $output): ?int
             return 0;
         }
 
-        if ($latestIndexExists && $force) {
+        if ($latestIndexExists) {
```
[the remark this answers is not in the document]
grubolsch: You don't need the force check because the piece of code before it:

```php
if ($latestIndexExists && !$force) {
            // Latest index already exists, do nothing.
            $logger->info('Latest geoshapes index exists already. Aborting installation.');
            return 0;
        }
```
If latestIndexExists = true, then we we know that force must be false after this if, otherwise it would have gone into the first if, and ended the function because of the return.
You could rewrite this to be an "else if", but that also would not be needed - because of the return statement in the first if.

app/Console/InstallUDB3CoreCommand.php
grubolsch: How is this valid PHP syntax? :_O

app/Console/InstallUDB3CoreCommand.php
grubolsch: Same story as above, because the if above it with a return it is not needed.

src/Http/Authentication/JsonWebToken.php
grubolsch: Really nice! It works, thank you!

tests/AMQP/Dummies/DummyEventNotSerializable.php
grubolsch: I get  

```
  Line   tests/AMQP/Dummies/DummyEventNotSerializable.php                                                               
  9      Property CultuurNet\UDB3\Search\AMQP\Dummies\DummyEventNotSerializable::$id is never read, only written.       
         💡 See: https://phpstan.org/developing-extensions/always-read-written-properties                                
  10     Property CultuurNet\UDB3\Search\AMQP\Dummies\DummyEventNotSerializable::$content is never read, only written.  
         💡 See: https://phpstan.org/developing-extensions/always-read-written-properties  
```                            

PHPstan detects it is a "useless" class because you could never access any of the properties once created.
It of course does not understand the context in which it is used: a unit test.

I would either keep it like this, or add a phpstan-ignore-file comment in the file.

What do you prefer?

LucWollants: It's a test class and the code is not really needed so I would use a `phpstan-ignore-file` or just make the properties public?

src/ElasticSearch/IndexationStrategy/BulkIndexationStrategy.php
grubolsch: The docblock was just incorrect. If you look in the `indexDocument()` method you can see it actually contains an assoc array.

```
public function indexDocument(
        string $indexName,
        string $documentType,
        JsonDocument $jsonDocument
    ): void {
        $id = $jsonDocument->getId();
        $this->logger->info("Queuing document {$id} for indexation.");

        $this->queuedDocuments[] = [
            'index' => $indexName,
            'type' => $documentType,
            'id' => $jsonDocument->getId(),
            'body' => Json::decodeAssociatively($jsonDocument->getRawBody()),
        ];

        $this->autoFlush();
    }
```

I actually started to refactor this to a new class to improve this, but I stopped because of 2 reasons:
- Out of scope of my current ticket - but would be happy to create a ticket later and pick it up
- I am not sure that maybe it NEEDS to be a simple array because of the context of this class, I think it has something to do with indexing documents for ES, and maybe it cannot handle more complex data structures? Not sure, just trying to guess based on this class.

src/JsonDocument/GuzzleJsonDocumentFetcher.php
grubolsch: Had to with problems with the interface JsonDocumentFetcher which returns static.
But I could change it to "self", which is fine because this is a final class.

tests/AMQP/Dummies/DummyEventNotSerializable.php
grubolsch: Made it public

LucWollants (approved):

simon-debruijn (approved):

PR #278, opened 2024-04-08
III-3818 Remove useless dependencies sapi3

grubolsch:
### Added
- Added composer-unused to project, and added it to the github build pipeline.
 
### Removed
 - removed unused package php-http/guzzle7-adapter
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3818

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,3 +105,34 @@ jobs:
 
             - name: 🔍 Run static analysis
               run: composer phpstan
+
+    unused:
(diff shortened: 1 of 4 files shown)


grubolsch:
Maybe we will not do this after all - because it even downgrades some packages and does not really find anything actionable.

LucWollants:
> Maybe we will not do this after all - because it even downgrades some packages and does not really find anything actionable.

Lets close it for now

PR #280, opened 2024-04-08
Refactor duplicate code, write extra unit test

grubolsch:
### Changed
 
- I saw there was duplicate code in SortByOfferRequestParser & SortByOrganizerRequestParser and decided to refactor it to use 1  shared method.
- I introduced 2 new unit tests for this new class, just for some sanity checks.

### DIFF
--- a/src/Http/Offer/RequestParser/QueryBuilderFactory.php
+++ b/src/Http/Offer/RequestParser/QueryBuilderFactory.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search\Http\Offer\RequestParser;
+
+use CultuurNet\UDB3\Search\Offer\OfferQueryBuilderInterface;
+use CultuurNet\UDB3\Search\Organizer\OrganizerQueryBuilderInterface;
+use CultuurNet\UDB3\Search\QueryBuilder;
+use CultuurNet\UDB3\Search\SortOrder;
+use CultuurNet\UDB3\Search\UnsupportedParameterValue;
+
+final class QueryBuilderFactory
+{
+    /** @return OrganizerQueryBuilderInterface|OfferQueryBuilderInterface */
+    public static function getQueryBuilder(array $sorts, array $sortBuilders, QueryBuilder $queryBuilder)
+    {
+        foreach ($sorts as $field => $order) {
+            if (!isset($sortBuilders[$field])) {
+                throw new UnsupportedParameterValue("Invalid sort field '{$field}' given.");
+            }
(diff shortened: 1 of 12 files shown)


src/Http/Offer/RequestParser/QueryBuilderFactory.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace CultuurNet\UDB3\Search\Http\Offer\RequestParser;
+
+use CultuurNet\UDB3\Search\Offer\OfferQueryBuilderInterface;
+use CultuurNet\UDB3\Search\Organizer\OrganizerQueryBuilderInterface;
+use CultuurNet\UDB3\Search\QueryBuilder;
+use CultuurNet\UDB3\Search\SortOrder;
+use CultuurNet\UDB3\Search\UnsupportedParameterValue;
+
+final class QueryBuilderFactory
```
LucWollants, line 13: Can this not become part of the already existing `QueryBuilder`?

grubolsch: QueryBuilder is an interface, not a class.
I could change it so both RequestParsers would extend the same abstract class, but I wanted to avoid introducing inheritance.

LucWollants: Indeed, it needs to be part of the concrete implementation.

But you created a very general class named `QueryBuilderFactory` with a very general method `getQueryBuilder` which validates the sort builders. So I would create something like `withSortBuilders` on the interface `QueryBuilder`

grubolsch: Not sure if I am following.
You want me to extend both classes with an abstract base class, and then add a new method to the interface?

tests/SortBuildersTest.php
```diff
@@ -36,6 +34,6 @@ public function testGetQueryBuilderUnsupportedSort(): void
 
         $this->expectException(UnsupportedParameterValue::class);
 
-        QueryBuilderFactory::getQueryBuilder($sorts, $sortBuilders, $queryBuilderMock);
+        (new SortBuilders())->withSortBuilders($sorts, $sortBuilders, $queryBuilderMock);
```
simon-debruijn, line 37: This looks very meta 😛

LucWollants: Maybe this is better
```
(new SortBuilders($queryBuilderMock))->build($sorts, $sortBuilders);
```

grubolsch:
All feedback has been processed

LucWollants (approved):

PR #290, opened 2024-06-17
III-6220 Refactor ManagementToken as preparation for Keycloak

LucWollants:
### Changed
- Refactor `Auth0` specific management code to interfaces so they can be implemented with a Keycloak flavor
- The `JsonDocumentFetcher` only needs a token
- The `AuthenticateRequest` also needs to make a management call

### Note
- The behavior should be the same as before. But with these changes and interfaces, it is possible to implement a Keycloak flavor
 
---

Ticket: https://jira.uitdatabank.be/browse/III-6220


src/Http/Authentication/Auth0/Auth0ManagementTokenGenerator.php
```diff
@@ -0,0 +1,83 @@
...
+use CultuurNet\UDB3\Search\Json;
+use DateTimeImmutable;
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\ConnectException;
+use GuzzleHttp\Psr7\Request;
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerAwareTrait;
+use Psr\Log\NullLogger;
+
+final class Auth0ManagementTokenGenerator implements ManagementTokenGenerator, LoggerAwareInterface
```
grubolsch, line 18: Is there any way to set the Logger? It should be an optional dependency?

src/Http/Authentication/Auth0/Auth0ManagementTokenGenerator.php
```diff
@@ -0,0 +1,83 @@
...
+    ) {
+        $this->domain = $domain;
+        $this->client = $client;
+        $this->clientId = $clientId;
+        $this->clientSecret = $clientSecret;
+        $this->audience = $audience;
+        $this->logger = new NullLogger();
+    }
+
+    public function newToken(): ManagementToken
```
grubolsch, line 47: I personally like the name functions that make api calls with the prefix "fetch". This fetches a different token each time you call it, it does not return the same "new token".
So I would rename it to fetchNewToken()

src/Http/Authentication/ManagementToken/ManagementTokenProvider.php
```diff
@@ -0,0 +1,39 @@
...
+    ) {
+        $this->tokenGenerator = $tokenGenerator;
+        $this->tokenRepository = $tokenRepository;
+    }
+
+    public function token(): string
+    {
+        $token = $this->tokenRepository->get();
+
+        if ($token === null || $this->expiresWithin($token, '+5 minutes')) {
```
grubolsch, line 27: Move string to a constant

src/Http/Authentication/ManagementToken/ManagementTokenProvider.php
```diff
@@ -0,0 +1,39 @@
...
+            $token = $this->tokenGenerator->newToken();
+            $this->tokenRepository->set($token);
+        }
+
+        return $token->getToken();
+    }
+
+    private function expiresWithin(ManagementToken $token, string $offset): bool
```
grubolsch, line 35: offset will always be +5 minutes so remove param and use constant/string hardcoded?

src/Http/Authentication/ManagementToken/ManagementTokenFileRepository.php
```diff
@@ -16,22 +16,22 @@ public function __construct(string $fullFilePath)
...
+    public function get(): ?ManagementToken
     {
         if (!file_exists($this->fullFilePath)) {
             return null;
         }
 
         $tokenAsArray = Json::decodeAssociatively(file_get_contents($this->fullFilePath));
 
-        return new Auth0Token(
+        return new ManagementToken(
```
grubolsch, line 27: Should we check if the fields `token`, `issuesAt` & `expiresIn` are returned? If not return `NULL`?

src/Http/Authentication/Auth0/Auth0ManagementTokenGenerator.php
```diff
@@ -0,0 +1,83 @@
...
+                throw new ConnectException(
+                    $message,
+                    new Request('POST', 'https://' . $this->domain . '/oauth/token')
+                );
+            }
+
+            $this->logger->info($message);
+        }
+
+        $res = Json::decodeAssociatively($response->getBody()->getContents());
```
grubolsch, line 76: I think this is a difference in style, but I would check if the fields `access_token` & `expires_in` are returned.
And check if JSON throws a JsonException, in either of these cases I would throw a ConnectException.

grubolsch (changes requested):

src/Http/Authentication/Auth0/Auth0ManagementTokenGenerator.php
LucWollants: It works with the `LoggerAwareTrait` trait, but it was never set before. We could create a follow-up ticket to change this behavior.

src/Http/Authentication/Auth0/Auth0ManagementTokenGenerator.php
LucWollants: The method comes from the interface `ManagementTokenGenerator`, this interface does not imply that the token is created through an HTTP request, the token could also be created locally. Maybe `generate` is a better name that matches the `ManagementTokenGenerator` interface better?

src/Http/Authentication/ManagementToken/ManagementTokenProvider.php
LucWollants: I changed the method to `isExpired` and the offset is internal, so no longer need for a constant

src/Http/Authentication/Auth0/Auth0ManagementTokenGenerator.php
LucWollants: That means there was something wrong with the `set` method and we would never notice this. If we overlook that and always return `null` from the cache we will always create a new token. So I would prefer a hard crash that allows us to fix the issue.

src/Http/Authentication/ManagementToken/ManagementTokenFileRepository.php
LucWollants: That means there was something wrong with the set method and we would never notice this. If we overlook that and always return null from the cache we will always create a new token. So I would prefer a hard crash that allows us to fix the issue.

src/Http/Authentication/Auth0/Auth0ManagementTokenGenerator.php
grubolsch: I can live with generate :-)

src/Http/Authentication/ManagementToken/ManagementTokenFileRepository.php
grubolsch: okay

grubolsch (approved):

src/Http/Authentication/Auth0/Auth0ManagementTokenGenerator.php
LucWollants: I became a little more complex. Turns out we need to create a management token to validate an incoming access token an we need to create a login token to make authenticated requests to UDB3. So in the follow-up pull request the interface changed again 😅

PR #307, opened 2024-08-20
III-3558 Improve `file_get_contents` return value

LucWollants:
### Changed
- Improved `file_get_contents` return value by adding a wrapper method which always returns a string. In case of an error, it will not return false, but throw an exception. Note that all files are internally managed so if something is wrong we want to know this fast, so an exception is much better than an ignored return value.
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3558


app/RoutingServiceProvider.php
```diff
@@ -63,7 +64,7 @@ function (): Router {
                         new InMemoryDefaultQueryRepository(
                             file_exists(__DIR__ . '/../default_queries.php') ? require __DIR__ . '/../default_queries.php' : []
                         ),
-                        file_get_contents('file://' . __DIR__ . '/../' . $pemFile)
+                        FileReader::read('file://' . __DIR__ . '/../' . $pemFile)
```
simon-debruijn, line 67: For this case alone, this might be overkill but...

It could be useful to depend on known apis when doing something as frequent as reading files.

A php developer will know how https://symfony.com/doc/current/components/filesystem.html#readfile behaves for example.
Whereas if we create custom wrappers for several basic things, it might not be clear to use our wrapper instead

simon-debruijn (approved):

JonasVHG (approved):

PR #310, opened 2024-08-23
III-3558 Increase PHPStan to level 8

LucWollants:
### Changed
- Increased PHPStan to level 8 by mainly fixing nullable types
 
---

Ticket: https://jira.uitdatabank.be/browse/III-3558


src/Http/OrganizerSearchController.php
```diff
@@ -67,9 +67,11 @@ public function __construct(
 
     public function __invoke(ApiRequestInterface $request): ResponseInterface
     {
-        $this->organizerParameterWhiteList->guardAgainstUnsupportedParameters(
-            $request->getQueryParamsKeys()
-        );
+        $keys = $request->getQueryParamsKeys();
+
+        if ($keys) {
```
simon-debruijn, line 72: maybe this if can be a bit more specific?
It is to check !== null I presume

simon-debruijn (approved):

app/Console/AbstractCommand.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\SearchService\Console;
+
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Command\Command;
+
+abstract class AbstractCommand extends Command
```
JonasVHG, line 10: This kinda goes against our policy not to use prefixes & suffixes with `Abstract`, 
but the alternatives I've tried won't win any beauty contests either 🤐, so maybe best leave it like this.

JonasVHG (approved):

app/Console/AbstractCommand.php
LucWollants: Maybe we need to create an extra ticket to enforce these code guidelines? With the PHP 8.x upgrade in mind we need to run Rector to improve the code base. One of the custom rules could be to avoid this kind of suffixes and prefixes.

But for now I will not change it, it's also not in the scope of this ticket and pull request.

PR #324, opened 2024-10-18
Add healthcheck to PHP container

willaerk:
---

Ticket: https://jira.uitdatabank.be/browse/III-6154

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,9 +12,16 @@ services:
       uitdatabank-search:
         aliases:
           - search.uitdatabank.local
+    healthcheck:
+      test: curl http://127.0.0.1:80/
+      start_period: 5s
+      interval: 5s
+      timeout: 5s
+      retries: 10
     depends_on:
       elasticsearch:
         condition: service_healthy
+
   elasticsearch:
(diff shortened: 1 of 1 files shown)


JonasVHG (approved):

docker-compose.yml
```diff
@@ -12,9 +12,16 @@ services:
       uitdatabank-search:
         aliases:
           - search.uitdatabank.local
+    healthcheck:
+      test: curl http://127.0.0.1:80/
```
LucWollants, line 16: Is this not the wrong port? Or it needs to be the internal port?

willaerk: It runs inside the container, so it needs to be the regular listen port of the webserver process. The healthcheck is used to be able to mark the container as "healthy" after starting it.

willaerk: The health checks are especially handy to create container dependencies, in our case for example to make sure the MySQL service is running and listening for connections before trying to run the database migrations.
For the php container, it is useful to avoid having the next step (the feature-init step) run before the webserver is up.

LucWollants: Ok, I see! Indeed it needs to run inside the container, that makes sense. Thanks for the clear explanation.

PR #328, opened 2024-11-18
III-6410 Cleanup auth0 Code

JonasVHG:
### Changed
 
- `app/RoutingServiceProvider`: `pemfile` will only be the `Keycloak` one & `getManagementTokenProvider()` will always be a `Keycloak` one.
- `JsonDocumentFetcherProvider`: `getTokenGenerator()` will always return a `KeycloakTokenGenerator`
- `Authentication/AuthenticateRequest`: rename `$auth0Down` to `$oAuthServerDown ` & `$jwtUrl` is always set.
- `AuthenticateRequestTest`: Use `KeycloakMetadataGenerator` instead of `Auth0MetadataGenerator` in tests & update MockResponses with Keycloak metadata.
 
### Removed
 
- `Auth0TokenGenerator`
- `Auth0MetadataGenerator`

### Related PR

- https://github.com/cultuurnet/appconfig/pull/891
 
---

Ticket: https://jira.publiq.be/browse/III-6410


src/Http/Authentication/AuthenticateRequest.php
```diff
@@ -92,7 +92,7 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
 
     private function handleClientId(ServerRequestInterface $request, RequestHandlerInterface $handler, string $clientId): ResponseInterface
     {
-        $auth0Down = false;
+        $oauthServerDown = false;
```
LucWollants, line 95: I wonder, should this be `$oAuthServerDown` ?

LucWollants (changes requested):
I think this can also be changed because `keycloak` will always be enabled https://github.com/cultuurnet/udb3-search-service/blob/86e610afa138a5ac1af959eebc30a99d10fc6530/src/Http/Authentication/AuthenticateRequest.php#L148

LucWollants:
Maybe prepare a config change that removes all the Auth0-related config? However, we need to wait to merge that cleanup of the config until this code is deployed to production.

src/Http/Authentication/AuthenticateRequest.php
JonasVHG: Indeed, that might improve readability.

JonasVHG:
> Maybe prepare a config change that removes all the Auth0-related config? However, we need to wait to merge that cleanup of the config until this code is deployed to production.

Yes, I decided to wait with this task, because as long it is not deployed it cannot be merged. I could also make one with a `wait-to-merge`-label.

LucWollants:
> Yes, I decided to wait with this task, because as long it is not deployed it cannot be merged. I could also make one with a `wait-to-merge`-label.

I would already make a pull request, so we can put some "pressure" on releasing everything

PR #331, opened 2025-01-17
Added unused packages detector package

Anahkiasen:
### Added
 
- Added unused packages detector package
 
---

Only `php-http/guzzle7-adapter` is unused, unfortunately removing it is impossible because the `php-http/discovery` plugin automatically adds it back to ensure we have a valid PSR implementation. But since Guzzle does already come with one, I'm guessing this should not trigger? Or were we making Guzzle interact with a non PSR18 library somewhere, I didn't see that in the Guzzle parts but maybe I missed it 🤔 If not I can perhaps turn off the plugin discovery

Ticket: https://jira.publiq.be/browse/III-3818

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,4 +135,4 @@ jobs:
           run: composer install --no-progress --no-suggest
 
         - name: 🗑 Run unused dependencies check
-          run: composer unused
+          run: vendor/bin/composer-unused
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,7 @@ jobs:
       strategy:
         matrix:
           php-versions: ['8.1']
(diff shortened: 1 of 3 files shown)


LucWollants:
I see we don't have `php-http/discovery` inside https://github.com/cultuurnet/udb3-backend/blob/master/composer.json

Anahkiasen:
@LucWollants Are you saying I should add it to `backend` or remove it here?

LucWollants:
> @LucWollants Are you saying I should add it to `backend` or remove it here?

No, I guess that was the solution in `udb3-backend` to make the unused package work?

Anahkiasen:
Okay will do the same here too then, I don't think the check is working properly for some reason

Anahkiasen:
Done for backend in https://github.com/cultuurnet/udb3-backend/pull/1961

LucWollants (approved):

PR #332, opened 2025-01-21
III-6474 Add Redis Cache to Search

JonasVHG:
### Added
 
- `CacheProvider`: new ServiceProvider for caches
 
### Changed
 
- `composer.json`: Add `symfony/cache`  & `predis/predis`
- `ContainerFactory`: Register new ServiceProvider with flag.

### Related PR

- https://github.com/cultuurnet/appconfig/pull/938
 
---

Ticket: https://jira.publiq.be/browse/III-6474


app/CacheProvider.php
```diff
@@ -0,0 +1,30 @@
...
+            RedisAdapter::class,
+            fn (): RedisAdapter =>
+            new RedisAdapter(
+                new Client(
+                    $this->parameter('cache.redis')
+                ),
+                'permission' . '_',
+                86400,
+            ),
+        );
```
simon-debruijn, line 28:
```suggestion
    protected $provides = [
        CacheInterface::class,
    ];

    public function register(): void
    {
        $this->add(
            CacheInterface::class,
            fn (): RedisAdapter =>
            new RedisAdapter(
                new Client(
                    $this->parameter('cache.redis')
                ),
                'permission' . '_',
                86400,
            ),
        );
```

it could make sense to depend on the interface that the adapters adhere to. So that you could swap out the implementation.

simon-debruijn (changes requested):

app/Factory/ContainerFactory.php
```diff
@@ -39,6 +40,9 @@ public static function forCli(Config $config): Container
         $container->addServiceProvider(EventIndexationServiceProvider::class);
         $container->addServiceProvider(PlaceIndexationServiceProvider::class);
         $container->addServiceProvider(CommandServiceProvider::class);
+        if ($container->get(Config::class)->get('cache.enabled') ?? false) {
```
grubolsch, line 43: Not sure I like the idea of a service that can exist or not exist depending on config.
This can lead easy to crashes when a service is not there.

I would make a CacheProvider that just does nothing, but implements and interface like CacheProvider, so no matter if caching is enabled or "some" class exists for it.

grubolsch (changes requested):

app/Factory/ContainerFactory.php
JonasVHG: That is also an idea. Since this is a temporary feature flag, till search has a redis instace, I didn't weigh out all the pros & cons.

LucWollants: I prefer a hard crash here, what is the use of muting a critical error? You would never know if the cache is missing but you expected it to exist.

LucWollants (approved):

app/CacheProvider.php
LucWollants: This was fixed, can you approve @simon-debruijn

app/Factory/ContainerFactory.php
LucWollants: @JonasVHG and @grubolsch if needed we could quickly discuss this, otherwise I would prefer to try it

simon-debruijn (approved):

grubolsch (approved):

PR #333, opened 2025-01-29
III-6472 cache integrations

JonasVHG:
### Changed
 
- `AuthenticateRequest`: Add `RedisAdaptor` & save `clientId` & `apiKey` in redisCache
- `RoutingServiceProvider`: Update constructor of AuthenticateRequest
- `AuthenticateRequestTest`: Update existing Unit tests & add new Unit tests for cached `apiKeys` & `clientIds`

### Fixed

- `apiKeys` & `clientIds`:  Can now be cached.
 
---

### DIFF
--- a/src/Http/Authentication/AuthenticateRequest.php
+++ b/src/Http/Authentication/AuthenticateRequest.php
@@ -27,6 +27,8 @@ use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
 use Psr\Log\LoggerAwareInterface;
 use Psr\Log\NullLogger;
+use Symfony\Component\Cache\Adapter\RedisAdapter;
+use Symfony\Component\Cache\CacheItem;
 
 final class AuthenticateRequest implements MiddlewareInterface, LoggerAwareInterface
 {
@@ -46,13 +48,16 @@ final class AuthenticateRequest implements MiddlewareInterface, LoggerAwareInter
 
     private string $pemFile;
 
+    private RedisAdapter $redisCache;
+
     public function __construct(
         Container $container,
         ICultureFeed $cultureFeed,
         ManagementTokenProvider $managementTokenProvider,
         MetadataGenerator $metadataGenerator,
         DefaultQueryRepository $defaultQueryRepository,
-        string $pemFile
+        string $pemFile,
+        RedisAdapter $redisCache
     ) {
         $this->container = $container;
         $this->cultureFeed = $cultureFeed;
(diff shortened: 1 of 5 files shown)


src/Http/Authentication/AuthenticateRequest.php
```diff
@@ -46,20 +48,24 @@ final class AuthenticateRequest implements MiddlewareInterface, LoggerAwareInter
...
+
     public function __construct(
         Container $container,
         ICultureFeed $cultureFeed,
         ManagementTokenProvider $managementTokenProvider,
         MetadataGenerator $metadataGenerator,
         DefaultQueryRepository $defaultQueryRepository,
-        string $pemFile
+        string $pemFile,
+        RedisAdapter $redisCache
```
LucWollants, line 60: It feels rather weird that a concrete `RedisAdapter` is injected here. I would expect a `CacheInterface` or another interface to be injected here. The actual code to set and get keys should not be implemented inside `AuthenticateRequest` class but in a new `AccessKeyCache` (or other name)

src/Http/Authentication/AuthenticateRequest.php
```diff
@@ -93,24 +99,29 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+                );
+
+                if ($metadata === null) {
+                    return (new InvalidClientId($clientId))->toResponse();
+                }
+                $hasSapiAccess->set($this->hasSapiAccess($metadata));
+                $this->redisCache->save($hasSapiAccess);
+            } catch (ConnectException $connectException) {
+                $this->logger->error('OAuth server was detected as down, this results in disabling authentication');
+                $oAuthServerDown = true;
```
grubolsch, line 119: $this->logger->error('OAuth server was detected as down, this results in disabling authentication');
            
I know that this was already the case, but am I the only one that finds this very scary?
This means if a hacker ever ddoss or otherwise disables Keycloak, we just open our data to everybody?

JonasVHG: This is read-only data, not writing. And we have contracts with big commercial parties. Do I do not know how realistic it is to change this.

LucWollants: The amount of data that is returned is still limited to the query and the pagination of Elastic Search. So the reasoning behind this was: "The inconvenience for a normal user is worse than the public data a hacker could potentially steal"

LucWollants:
As discussed through Slack a cleaner approach could be to use caching decorators after extracting the code that interacts with CultureFeed and the code that interacts with Keycloak.

JonasVHG:
See https://github.com/cultuurnet/udb3-search-service/pull/334 for actual solution

PR #336, opened 2025-02-03
III-6472 Add CachedClientIdProvider

JonasVHG:
### Added
 
- `CachedClientIdResolver`: Add `ClientIdProvider` with Caching 
- `CachedClientIdResolverTest`: Unit tests
 
### Changed
 
- `RoutingServiceProvider`: Add new instance of `CachedClientIdResolver` & use in the creation of `AuthenticateRequest` instead of `MetadataClientIdProvider`
 
---


src/Http/Authentication/Access/CachedClientIdProvider.php
```diff
@@ -0,0 +1,34 @@
...
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search\Http\Authentication\Access;
+
+use Symfony\Component\Cache\Adapter\AbstractAdapter;
+use Symfony\Component\Cache\CacheItem;
+
+final class CachedClientIdProvider implements ClientIdProvider
+{
+    private AbstractAdapter $cache;
```
LucWollants, line 12: Should this not be something like `CacheItemInterface` or a `CacheInterface` of that exists?

tests/Http/Authentication/Access/CachedClientIdProviderTest.php
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * @var ClientIdProvider&MockObject
+     */
+    private $clientIdProvider;
+
+    /**
+     * @var CacheItemInterface&MockObject
+     */
+    private $cacheItem;
```
LucWollants, line 27: I wonder if we could work here with a simple adapter implementation (something like an array flavour) to closer match the actual behaviour?

src/Http/Authentication/Access/CachedClientIdProvider.php
LucWollants: But `CacheInterface` has the method `get`

grubolsch (approved):

tests/Http/Authentication/Access/CachedClientIdProviderTest.php
```diff
@@ -27,18 +28,21 @@ protected function setUp(): void
...
     public function it_will_use_cached_values(bool $hasAccess): void
     {
+        $cache = new ArrayAdapter();
+        $cache->get(
+            'my_cached_client_id',
+            function () use ($hasAccess) {
+                return $hasAccess;
+            }
+        );
         $this->cachedClientIdProvider = new CachedClientIdProvider(
```
LucWollants: I think this can be moved to the `setup` method

LucWollants (approved):

PR #338, opened 2025-02-04
III-6472 Add CachedConsumerProvider

JonasVHG:
### Added
 
- `CachedConsumerResolver`: `ConsumerResolver` with caching
- `CachedConsumerResolverTest`: Unit tests for `CachedConsumerResolver`
 
### Changed
 
- `RoutingServiceProvider`: Use `CachedConsumerResolver` instead of `CulturefeedConsumerResolver` if toggle is enabled
- `CachedClientIdResolver`: Use suffixes for clarity
- `CachedClientIdResolverTest`: Use suffixes for clarity + refactored unnecessary code in existing unit test.
 
---

### DIFF
--- a/src/Http/Authentication/Access/CachedClientIdResolver.php
+++ b/src/Http/Authentication/Access/CachedClientIdResolver.php
@@ -22,10 +22,15 @@ final class CachedClientIdResolver implements ClientIdResolver
     public function hasSapiAccess(string $clientId): bool
     {
         return $this->cache->get(
-            $clientId,
+            $this->addSuffix($clientId),
             function () use ($clientId) {
                 return $this->clientIdAccess->hasSapiAccess($clientId);
             }
         );
+    }
+
+    private function addSuffix(string $clientId): string
+    {
+        return 'client_id_' . $clientId . '_sapi_access';
     }
 }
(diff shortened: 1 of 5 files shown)


src/Http/Authentication/Access/CachedClientIdResolver.php
```diff
@@ -22,10 +22,15 @@ public function __construct(
...
         return $this->cache->get(
-            $clientId,
+            $this->addSuffix($clientId),
             function () use ($clientId) {
                 return $this->clientIdAccess->hasSapiAccess($clientId);
             }
         );
     }
+
+    private function addSuffix(string $clientId): string
```
LucWollants, line 32: Maybe a better name is `createKey`

tests/Http/Authentication/Access/CachedConsumerResolverTest.php
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * @test
+     */
+    public function it_will_get_the_cached_status_value(): void
+    {
+        $this->consumerResolver->expects($this->never())
+            ->method('getStatus')
+            ->with('my_cached_api_key')
+            ->willReturn('ACTIVE');
```
LucWollants, line 51: I'm a bit confused how a method that never gets called can return something?

src/Http/Authentication/Access/CachedClientIdResolver.php
JonasVHG: I've used `createCacheKey()` because we also have `apiKeys`, it might be confusing at first glance.

LucWollants (approved):

PR #344, opened 2025-02-12
III-6519 redis mgmt token cache

steftrenson:
### Changed 
- `ManagementTokenProvider` now uses a redis-cache-based repository instead of a file.

---

Ticket: https://jira.publiq.be/browse/III-6519


app/RoutingServiceProvider.php
```diff
@@ -157,7 +165,13 @@ private function getManagementTokenProvider(): ManagementTokenProvider
                 $this->parameter('keycloak.client_secret'),
                 $this->parameter('keycloak.domain') . '/api/v2/'
             ),
-            new ManagementTokenFileRepository(__DIR__ . '/../cache/keycloak-management-token-cache.json'),
+            new CacheBasedManagementTokenRepository(
+                CacheFactory::create(
+                    $this->getContainer(),
+                    '',
```
steftrenson, line 171: I'm wondering if this should have a namespace, but I don't know if it would add any value.

cache/.gitignore
```diff
@@ -1,4 +0,0 @@
-# Ignore everything in this directory
```
steftrenson, line 1: Can someone double check this cache directory is actually not used now anymore? It was used to store the file for the mgmt token and it looks like that was the only place but some verification would be nice.

src/Cache/CacheFactory.php
```diff
@@ -0,0 +1,21 @@
...
+
+use Predis\Client;
+use Psr\Container\ContainerInterface;
+use Symfony\Component\Cache\Adapter\RedisAdapter;
+use Symfony\Contracts\Cache\CacheInterface;
+
+final class CacheFactory
+{
+    public static function create(
+        ContainerInterface $container,
```
LucWollants, line 15: To make it more explicit, you could pass the `Predis\Client` here directly?

app/RoutingServiceProvider.php
LucWollants, line 171: Nit: should we also give this a namespace?

cache/.gitignore
JonasVHG: The folder was indeed only used for storing the management token(in the past for `auth0` now for `keycloak`

LucWollants: Makes me wonder if this could have an impact on puppet code @willaerk?

LucWollants (approved):

cache/.gitignore
LucWollants: As discussed during the biweekly meeting:
- when doing a deployment delete the folder
- acceptance tests could have an issue

JonasVHG (dismissed):

JonasVHG (approved):

PR #439, opened 2026-04-07
III-4554 Get Json Terms from Taxonomy endpoint instead of hardcoded files.

JonasVHG:
### Added
 
- `TaxonomyApiClient`: Interface to handle `taxonomy`-stuff.
- `JsonTaxonomyApiClient`: Get taxonomy terms from Json Endpoint.
- `JsonTaxonomyApiClientTest`: Unit tests for `JsonTaxonomyApiClient`.
- `CachedTaxonomyApiClient`: Cache taxonomy from Json Endpoint.
- `CachedTaxonomyApiClientTest`: Unit tests for `CachedTaxonomyApiClient`.
- `TaxonomyApiProblem`: Custom Exception for Taxonomy problems.
- `TaxonomyServiceProvider`: Wire taxonomy stuff.
 
### Changed
 
- `ContainerFactory`: Register new ServiceProvider
- `OfferSearchServiceProvider`: Use terms from api endpoint instead of hardcoded from file.

### Related PR
 
- https://github.com/cultuurnet/appconfig/pull/1341
 
---

Ticket: https://jira.publiq.be/browse/III-4554

### DIFF
--- a/app/Taxonomy/TaxonomyServiceProvider.php
+++ b/app/Taxonomy/TaxonomyServiceProvider.php
@@ -4,12 +4,15 @@ declare(strict_types=1);
 
 namespace CultuurNet\UDB3\SearchService\Taxonomy;
 
+use CultuurNet\UDB3\Search\Cache\CacheFactory;
+use CultuurNet\UDB3\Search\Taxonomy\CachedTaxonomyApiClient;
 use CultuurNet\UDB3\Search\Taxonomy\JsonTaxonomyApiClient;
 use CultuurNet\UDB3\Search\Taxonomy\TaxonomyApiClient;
 use CultuurNet\UDB3\SearchService\BaseServiceProvider;
 use CultuurNet\UDB3\SearchService\Error\LoggerFactory;
 use CultuurNet\UDB3\SearchService\Error\LoggerName;
 use GuzzleHttp\Client;
+use Predis\Client as PredisClient;
 
 final class TaxonomyServiceProvider extends BaseServiceProvider
 {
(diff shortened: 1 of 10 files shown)


JonasVHG:
@claude, review.

app/Taxonomy/TaxonomyServiceProvider.php
```diff
@@ -0,0 +1,44 @@
...
+            TaxonomyApiClient::class,
+            fn (): TaxonomyApiClient => new CachedTaxonomyApiClient(
+                CacheFactory::create(
+                    $this->container->get(PredisClient::class),
+                    'taxonomy',
+                    86400 // one day
+                ),
+                new JsonTaxonomyApiClient(
+                    new Client(),
+                    $this->parameter('taxonomy.terms'),
```
lucwollants, line 35: I see there is an associated pull request on appconfig to remove the old term files. Should there not be a pull request to add the URL for `taxonomy.terms`?

JonasVHG: Indeed my mistake, I'll update the other PR.

tests/Taxonomy/JsonTaxonomyApiClientTest.php
```diff
@@ -0,0 +1,244 @@
...
+final class JsonTaxonomyApiClientTest extends TestCase
+{
+    private ClientInterface&MockObject $httpClient;
+
+    private NullLogger $logger;
+
+    public function setUp(): void
+    {
+        $this->httpClient = $this->createMock(ClientInterface::class);
+        $this->logger = new NullLogger();
```
lucwollants: I would not use a `NullLogger` here because we really want to get an error so we have a Sentry error and we can closely monitor issues.

src/Taxonomy/JsonTaxonomyApiClient.php
```diff
@@ -0,0 +1,68 @@
...
+
+    private function getTermsByDomain(string $domain): array
+    {
+        $termsByDomain = [];
+        foreach ($this->terms as $term) {
+            if ($term['domain'] === $domain) {
+                $termsByDomain[$term['id']]['name'] = $term['name'];
+            }
+        }
+        return $termsByDomain;
```
lucwollants, line 75: I wonder if we also should log something here when the array is empty?

tests/Taxonomy/JsonTaxonomyApiClientTest.php
JonasVHG: This is the create in the unit test. I do not think we want this in sentry.
The constructed one in the application has loggin: https://github.com/cultuurnet/udb3-search-service/blob/7cf40b4129066a1a1007ac9c2a98875ff31c9127/app/Taxonomy/TaxonomyServiceProvider.php#L36

lucwollants: I should have clarified better. We want to test that we actually report errors because this logging is important to trigger Sentry errors.

lucwollants (approved):

PR #441, opened 2026-04-08
III-7111 Search Requests via `POST` body

JonasVHG:
### Added
 
- `PostRequestBodyMiddleware`: new Middleware to handle `POST` requests
- `PostRequestBodyMiddlewareTest`: Unit Tests for new Middleware.
- `BadRequest` & `UnsupportedMediaType`: New apiProblems which can happen during a `POST`-request.
 
### Changed
 
- `RoutingServiceProvider`: Wire `PostRequestBodyMiddleware` in.
 
### Related PR
 
- https://github.com/cultuurnet/udb3-backend/pull/2221: Acceptance tests
- https://github.com/cultuurnet/apidocs/pull/502
 
---

Ticket: https://jira.publiq.be/browse/III-7111


JonasVHG:
@claude, review

tests/Http/Middleware/PostRequestBodyMiddlewareTest.php
```diff
@@ -0,0 +1,171 @@
...
+            }))
+            ->willReturn(new Response());
+
+        $this->middleware->process($request, $handler);
+    }
+
+    /**
+     * @test
+     */
+    public function it_replaces_url_query_params_with_body_params_on_post(): void
```
lucwollants: Did we not agree to report this as an error?

tests/Http/Middleware/PostRequestBodyMiddlewareTest.php
```diff
@@ -0,0 +1,171 @@
...
+use CultuurNet\UDB3\Search\Json;
+use Fig\Http\Message\StatusCodeInterface;
+use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use Slim\Psr7\Factory\ServerRequestFactory;
+use Slim\Psr7\Factory\StreamFactory;
+use Slim\Psr7\Response;
+
+final class PostRequestBodyMiddlewareTest extends TestCase
```
lucwollants, line 16: This seems to lack an example when the POST body contains unsupported parameters or even a fully random string.

tests/Http/Middleware/PostRequestBodyMiddlewareTest.php
JonasVHG: I thought that in the meeting we decided to ignore url parameters but an error is also acceptable & maybe the test name is not that accurate.
Anyone either giving an error or ignoring url params are both fine for me, as long as we do not try to mix them, which is confusing both for developers & end users.

lucwollants: I think we agreed that it would be a good UX experience for integrators to get an error (in combination with also documenting this). Otherwise, we'd be silently ignoring either the URL parameters or the POST body, which isn't as clear as giving an error stating that POST requests only support a body.

JonasVHG: After discussing this in Slack, we decided to return an error to the end user.

tests/Http/Middleware/PostRequestBodyMiddlewareTest.php
JonasVHG: I doubt that would be a useful test, since the code just passes it to a Mock of the `RequestHandlerInterface`.
The actual `OfferSearchController` & `OrganizerSearchController`, which are used regardless whether a `GET` or `POST` is used, have tests for those scenarios.(and an Acceptance test for this scenario is available & more useful).

src/Http/Middleware/PostRequestBodyMiddleware.php
```diff
@@ -20,17 +19,16 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
         if (stripos($contentType, 'text/plain') === false) {
-            $apiProblem = new ApiProblem('Unsupported Media Type', 'https://api.publiq.be/probs/body/unsupported-media-type');
-            $apiProblem->setStatus(StatusCodeInterface::STATUS_UNSUPPORTED_MEDIA_TYPE);
-            $apiProblem->setDetail('POST requests require Content-Type text/plain.');
-
-            return ResponseFactory::apiProblem(
-                $apiProblem->asArray(),
-                $apiProblem->getStatus()
-            );
+            return (new UnsupportedMediaType())->toResponse();
```
lucwollants, line 31: 🤩

src/Http/Middleware/PostRequestBodyMiddleware.php
```diff
@@ -40,4 +40,12 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
 
         return $handler->handle($request);
     }
+
+    private function stripAuthenticationParams(array $queryParams): array
+    {
+        return array_diff_key(
```
lucwollants: I must be honest, I don't understand how this works 😅

bertramakers:
```suggestion
        // Use array_diff_key to create a diff between the given query params (keys) and the authentication- 
        // specific query params (keys). The result is either an empty list if no other query params are
        // included in the request, or a list of (unsupported) query params.
        // Since array_diff_key creates a diff based on array keys, we use array_flip to create an array with
        // authentication-specific keys like clientId and apiKey.
        // E.g. ['clientId', 'apiKey'] becomes ['clientId' => 0, 'apiKey' => 1] 
        // (the values don't really matter in this case)
        return array_diff_key(
```

bertramakers: @JonasVHG maybe you can make the code itself easier to read using `array_filter()`. For example:

```
return array_filter($queryParams, function ($paramName) {
  return !in_array($paramName, ['clientId', 'apiKey']);
}, ARRAY_FILTER_USE_KEY);
```

(Quickly scribbled here, not sure if the syntax is 100% correct so make sure to verify)

JonasVHG: Done, and while looking at it again, I thought it might be even more readable to have that function return a bool instead of an array.

lucwollants (approved):
Looks good!

PR #446, opened 2026-04-14
III-7121 filter childrenOnly unless client has scope

JonasVHG:
### Changed
 
- `KeycloakMetadataGenerator`: Add new scope.
- `ClientIdResolver`, `MetadataClientIdResolver` & `CachedClientIdResolver`: Add `hasBoaAccess()`.
- `Consumer`: Add variable for storing `BoaAccess`.
- `OfferQueryBuilderInterface`, `ElasticSearchOfferQueryBuilder` & `MockOfferQueryBuilder`: Add `withAudienceTypeExcludeFilter()`.
- `AuthenticateRequest`: Add check for Boa-access.
- `OfferSearchController`: Add check for access to Boa.
- `MetadataClientIdResolverTest`, `CachedClientIdResolverTest`, `AuthenticateRequestTest` & `OfferSearchControllerTest`: Refactored/updated unit tests.
 
### Notes
 
- This PR only handles BOA-Access for Integrations with the correct `scope`. Access to the BOA-events, where you are the creator will be handled in the next PR.
- This PR will also break some acceptance tests, this will be handled in a PR in `udb3-backend`.

### Related PR

- https://github.com/cultuurnet/udb3-search-service/pull/448
- https://github.com/cultuurnet/udb3-backend/pull/2223
- https://github.com/cultuurnet/appconfig/pull/1351

---

Ticket: https://jira.publiq.be/browse/III-7121

### DIFF
--- a/src/Http/Authentication/Access/MetadataClientIdResolver.php
+++ b/src/Http/Authentication/Access/MetadataClientIdResolver.php
@@ -28,33 +28,12 @@ final class MetadataClientIdResolver implements ClientIdResolver
 
     public function hasSapiAccess(string $clientId): bool
     {
-        $oAuthServerDown = false;
-        $metadata = [];
-
-        try {
-            $metadata = $this->fetchMetadata($clientId);
-        } catch (ConnectException $connectException) {
-            $this->logger->error('OAuth server was detected as down, this results in disabling authentication');
-            $oAuthServerDown = true;
-        }
-
-        if (!$oAuthServerDown && !$this->hasApiAccess($metadata, 'sapi')) {
-            return false;
-        }
-
-        return true;
+        return $this->hasApiAccess($clientId, 'sapi');
     }
 
     public function hasBoaAccess(string $clientId): bool
     {
-        try {
-            $metadata = $this->fetchMetadata($clientId);
-        } catch (ConnectException $connectException) {
-            $this->logger->error('OAuth server was detected as down, this results in disabling boa access');
-            return false;
-        }
-
-        return $this->hasApiAccess($metadata, 'boa');
+        return $this->hasApiAccess($clientId, 'boa');
     }
 
     /**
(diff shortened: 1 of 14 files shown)


JonasVHG:
@claude, review.

src/Http/Authentication/Access/MetadataClientIdResolver.php
```diff
@@ -27,42 +27,57 @@ public function __construct(
...
 
-        if (empty($metadata['publiq-apis'])) {
-            return false;
+            $apis = explode(' ', $metadata['publiq-apis']);
+            return in_array($api, $apis, true);
         }
 
-        $apis = explode(' ', $metadata['publiq-apis']);
-        return in_array('sapi', $apis, true);
+        return true;
```
lucwollants: Would it not be safer to have a return `false` here and inside the catch we do a `return true`?

So it could become something like this

```
    private function hasApiAccess(string $clientId, string $api): bool
    {
        try {
            $metadata = $this->fetchMetadata($clientId);
        } catch (ConnectException $connectException) {
            $this->logger->error('OAuth server was detected as down, this results in disabling authentication');
            return true;
        }

        if (empty($metadata)) {
            return false;
        }

        if (empty($metadata['publiq-apis'])) {
            return false;
        }

        $apis = explode(' ', $metadata['publiq-apis']);
        return in_array($api, $apis, true);
    }
```

tests/Http/OfferSearchControllerTest.php
```diff
@@ -1212,6 +1212,180 @@ public function it_should_add_the_default_query_of_the_api_consumer_if_they_have
         $controller->__invoke(new ApiRequest($request));
     }
 
+    /**
+     * @test
+     */
+    public function it_returns_nothing_when_childrenOnly_requested_without_boa(): void
```
lucwollants, line 1218: I wonder if we should return an authorisation error here instead of just returning zero results? Using the audience type `childrenOnly` is not allowed when the consumer has no BOA access.

JonasVHG: The problem you should still be able to find your own `childrenOnly`-events, and since childrenOnly is by default a hidden `audienceType`, it can only be found in search results with the correct filter.
So if we throw an authorisation error, it would never be possible to find `childrenOnly`-events created by your client.

lucwollants: You are right! I made a wrong assumption that to get your "own" events with `audienceType:childrenOnly`, you didn't have to pass the `audienceType` as a filter option.

That makes me wonder what would happen if you search for `audienceType:*`

JonasVHG: Then you should get the following audienceTypes: `everybody`, `members`, `education` & depending on your scope only the `childrenOnly` events your created or all the `childrenOnly` events.

lucwollants: Is that something we should put in an acceptance test?

lucwollants (approved):

tests/Http/OfferSearchControllerTest.php
JonasVHG: I wille add 1 for `audienceType:*`

PR #447, opened 2026-04-14
III-7121 filter childrenOnly unless creator

JonasVHG:
### Added
 
-
 
### Changed
 
-
 
### Removed
 
-
 
### Fixed
 
-
 
---

Ticket: https://jira.publiq.be/browse/III-7121


JonasVHG:
This is only usefull for jwt-providers which we want to phase out.

PR #450, opened 2026-04-20
III-7140: Add MD5 hash check to mapping files

grubolsch:
## Summary
- Adds MD5 hash constants to `SchemaVersions` for each Elasticsearch mapping file (`mapping_udb3_core.json`, `mapping_event.json`, `mapping_place.json`, `mapping_organizer.json`, `mapping_region.json`)
- Adds test assertions that compute the actual MD5 of each file at test time and compare it to the stored constant
- If a mapping file is changed without updating the corresponding hash (and version number), CI will fail with a descriptive error message indicating which constant needs to be updated

## Test plan
- [ ] All existing `SchemaVersionsTest` tests still pass
- [ ] Verify that modifying a mapping JSON file causes the corresponding test to fail with a clear message
- [ ] Verify that updating both the hash constant and version constant makes the test pass again

### DIFF
--- a/bin/generate-schema-hashes.php
+++ b/bin/generate-schema-hashes.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+require_once __DIR__ . '/../vendor/autoload.php';
+
+use CultuurNet\UDB3\Search\ElasticSearch\Operations\SchemaVersions;
+
+$mappingDir = __DIR__ . '/../src/ElasticSearch/Operations/json/';
+
+$mappings = [
+    'UDB3_CORE_MAPPING_HASH' => ['file' => 'mapping_udb3_core.json', 'version' => SchemaVersions::UDB3_CORE],
+    'EVENT_MAPPING_HASH'     => ['file' => 'mapping_event.json',     'version' => SchemaVersions::UDB3_CORE],
+    'PLACE_MAPPING_HASH'     => ['file' => 'mapping_place.json',     'version' => SchemaVersions::UDB3_CORE],
+    'ORGANIZER_MAPPING_HASH' => ['file' => 'mapping_organizer.json', 'version' => SchemaVersions::UDB3_CORE],
+    'REGION_MAPPING_HASH'    => ['file' => 'mapping_region.json',    'version' => SchemaVersions::GEOSHAPES],
+];
+
+foreach ($mappings as $constant => $mapping) {
+    $contents = file_get_contents($mappingDir . $mapping['file']);
+    $hash = md5($contents . $mapping['version']);
+    echo "public const {$constant} = '{$hash}';" . PHP_EOL;
+}
--- a/bin/generate-schema-hashes.php
+++ b/bin/generate-schema-hashes.php
@@ -2,22 +2,18 @@
 
 declare(strict_types=1);
 
-require_once __DIR__ . '/../vendor/autoload.php';
-
-use CultuurNet\UDB3\Search\ElasticSearch\Operations\SchemaVersions;
-
 $mappingDir = __DIR__ . '/../src/ElasticSearch/Operations/json/';
 
-$mappings = [
-    'UDB3_CORE_MAPPING_HASH' => ['file' => 'mapping_udb3_core.json', 'version' => SchemaVersions::UDB3_CORE],
-    'EVENT_MAPPING_HASH'     => ['file' => 'mapping_event.json',     'version' => SchemaVersions::UDB3_CORE],
-    'PLACE_MAPPING_HASH'     => ['file' => 'mapping_place.json',     'version' => SchemaVersions::UDB3_CORE],
-    'ORGANIZER_MAPPING_HASH' => ['file' => 'mapping_organizer.json', 'version' => SchemaVersions::UDB3_CORE],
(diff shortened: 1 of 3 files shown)


grubolsch:
@claude

tests/ElasticSearch/Operations/SchemaVersionsTest.php
```diff
@@ -31,11 +31,11 @@ public function it_has_the_latest_geoshapes_version(): void
      */
     public function it_has_a_matching_hash_for_udb3_core_mapping(): void
     {
```
JonasVHG: Basically all these tests, do the same. 
I think readability could be improved with a private function like:
```
    /**
     * @test
     */
    public function it_has_a_matching_hash_for_udb3_core_mapping(): void
    {
        $this->assertCorrectHash(
            SchemaVersions::UDB3_CORE_MAPPING_HASH,
            'mapping_udb3_core.json'
        );
    }

    private function assertCorrectHash(string $expectedHash, string $filename): void
    {
        $actualHash = md5_file(self::MAPPING_DIR . $filename);
        $this->assertNotFalse($actualHash, 'Could not read ' . $filename);
        $this->assertSame(
            $expectedHash,
            $actualHash,
             $filename . ' has changed. Update SchemaVersions::UDB3_CORE and SchemaVersions::???.');
    }
```

Or maybe with a dataprovider?

grubolsch: Good idea, I added a data provider.

bin/generate-schema-hashes.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+require_once __DIR__ . '/../vendor/autoload.php';
+
+use CultuurNet\UDB3\Search\ElasticSearch\Operations\SchemaVersions;
+
+$mappingDir = __DIR__ . '/../src/ElasticSearch/Operations/json/';
+
+$mappings = [
+    'UDB3_CORE_MAPPING_HASH' => ['file' => 'mapping_udb3_core.json', 'version' => SchemaVersions::UDB3_CORE],
```
lucwollants: Does it make sense to keep working with `UDB3_CORE` and `GEOSHAPES` version if we won't update them? Should we not change the way the reindex is triggered to just use the hash?

bin/generate-schema-hashes.php
lucwollants, line 1: Why is this not a normal console command like we already have several?

bertramakers:
@grubolsch I like the idea of hashing the schema files to automatically check if the index schema is up-to-date or not, it's super clever and could really help prevent us from forgetting to update the schema version number! 💡 I love these kind of [Poka Yoke](https://nl.wikipedia.org/wiki/Poka_yoke) solutions.

However, the current approach seems to imply that we need to run the script to calculate the hashes manually, and then copy-paste the new versions to the SchemaVersions enum? If that's the case, I feel like we're close to a great improvement but _just_ missing the mark. Because we can still forget to run the script, we can make a mistake copy-pasting the output, etc ... Basically, if we can reduce the human involvement even further, we could prevent even more mistakes.

Would it be possible instead to remove the SchemaVersions enum, and replace it with an actual class that can calculate the right name of an index on the fly when we run the commands that need this info? It's probably relatively expensive if we'd always do it per incoming HTTP request, but since we use read and write aliases we should only need the actual index name with the hash in a couple of CLI commands so I don't think it would hurt performance-wise to always recalculate them in the few scenario's where we need them? Then we wouldn't need to remember to run the script whenever we make schema changes, copy-paste the output, commit the changes, make a PR, merge the PR, ...

Curious what you think!

Edit: I just realized you added a test to detect and warn us about new schema changes, also clever! But I still wonder if it wouldn't be easier / more efficient in the long run to just get rid of the SchemaVersions enum and always calculate the hashes on the fly whenever needed. 🤔  Because even with the test it still feels like it's a combination of quite a couple of moving pieces that could potentially break at some point.

grubolsch:
I now use the new approach of calculating the hash on the file as recommended by Bert

JonasVHG:
Clever solution!. But since this will affect infra, it would be best to check with @willaerk before merging.

JonasVHG (approved):

src/ElasticSearch/Operations/SchemaVersions.php
```diff
@@ -6,6 +6,29 @@
...
-    public const GEOSHAPES = 20250101000000;
+    private const MAPPING_DIR = __DIR__ . '/json/';
+
+    public static function udb3Core(): string
+    {
+        return md5(
+            self::readMappingFile('mapping_udb3_core.json') .
+            self::readMappingFile('mapping_event.json') .
+            self::readMappingFile('mapping_place.json') .
+            self::readMappingFile('mapping_organizer.json')
```
lucwollants: @grubolsch these are now legacy files. I don't think they even contain the new BOA fields.

lucwollants: Even the name of your method indicates we only need the new UDB3 core.

grubolsch:
Before merging, I will check with @willaerk if this needs infra changes.
Also to be used in combination with https://github.com/cultuurnet/udb3-search-service/pull/550

lucwollants (approved):
I noticed there is still a reference in the README to update the schema version which is no longer needed now, see https://github.com/cultuurnet/udb3-search-service/blob/main/readme.md?plain=1#L78

Something small to fix, so I already approved.

PR #455, opened 2026-05-06
III-7096 Make shard and replicas configurable

steftrenson:
### Changed
- Make Elasticsearch index shard and replica count configurable via `elasticsearch.number_of_shards` and `elasticsearch.number_of_replicas` config parameters, passed through `CreateIndexCommand` to the `CreateIndex` operation

---
Ticket: https://jira.uitdatabank.be/browse/III-7096

### DIFF
--- a/app/CommandServiceProvider.php
+++ b/app/CommandServiceProvider.php
@@ -81,6 +81,15 @@ final class CommandServiceProvider extends BaseServiceProvider
         );
 
         $this->add(
+            CreateIndexCommand::class,
+            fn (): CreateIndexCommand => new CreateIndexCommand(
+                $this->get(Client::class),
+                $this->parameter('elasticsearch.number_of_shards'),
+                $this->parameter('elasticsearch.number_of_replicas')
+            )
+        );
+
+        $this->add(
             UpdateUdb3CoreMappingCommand::class,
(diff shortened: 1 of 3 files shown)


app/CommandServiceProvider.php
```diff
@@ -80,6 +80,15 @@ function (): Application {
             }
         );
 
+        $this->add(
+            CreateIndexCommand::class,
+            fn (): CreateIndexCommand => new CreateIndexCommand(
+                $this->get(Client::class),
+                $this->parameter('elasticsearch.number_of_shards'),
```
lucwollants, line 87: What will be returned if these new config parameters are not yet defined? What is the default value they get, or will it crash?

lucwollants (approved):
I already approved, but I guess this needs an associated pull request on appconfig to set the numbers?

app/CommandServiceProvider.php
steftrenson: When the parameters are not available in the config file, these values will be null. When creating the index, it will then take the default values from ES itself, as they won't be added to the custom config array.

steftrenson:
@lucwollants The appconfig is already fixed in [this open PR](https://github.com/cultuurnet/appconfig/pull/1370/changes).

lucwollants (approved):
Looks good!

PR #460, opened 2026-05-20
III-7189 search api birthdate Advanced Query

JonasVHG:
### Added
 
- `BirthdateRangeTransformer.php`: Add transformer for `birthdateRange`
 
### Changed
 
- `mapping_event.json`: Add mapping for `birthdateRange`.
- `mapping_udb3_core.json`: Add mapping for `birthdateRange`.
- `SchemaVersions`: Update version.
- `EventTransformer`: Add new Transformer.
- `LuceneQueryStringFactory`: Add rewrite for `birthdateRange`.
- `LuceneQueryStringFactoryTest`: Unit tests
 
### Related PR
 
- https://github.com/cultuurnet/udb3-backend/pull/2247
- https://github.com/cultuurnet/apidocs/pull/523
 
### Note
 
- The url Parameter query & intersects with `typicalAgeRange` will be handled in the next PR.
 
---

Ticket: https://jira.publiq.be/browse/III-7189


JonasVHG:
@claude , review.

src/ElasticSearch/LuceneQueryStringFactory.php
```diff
@@ -12,6 +12,15 @@ final class LuceneQueryStringFactory implements QueryStringFactory
 
     public function fromString(string $queryString): LuceneQueryString
     {
+        // Rewrite shorthand date ranges `YYYY-MM-DD..YYYY-MM-DD` to the
+        // standard Lucene range syntax `[YYYY-MM-DD TO YYYY-MM-DD]`, so they
+        // can be used inside `q` queries (incl. within OR-groups).
+        $queryString = preg_replace(
```
lucwollants: I would not support this

tests/ElasticSearch/JsonDocument/Properties/BirthdateRangeTransformerTest.php
```diff
@@ -0,0 +1,129 @@
...
+            ],
+        ];
+
+        $this->assertEquals($expected, $this->transformer->transform($from, []));
+    }
+
+    /**
+     * @test
+     */
+    public function it_maps_only_from_when_to_is_absent(): void
```
lucwollants: Can this even happen? Because if I remember correctly we require both `from` and `to`

JonasVHG:
@claude, review.

src/ElasticSearch/Operations/json/mapping_udb3_core.json
```diff
@@ -226,6 +226,10 @@
             "type": "date",
             "format": "strict_date_time_no_millis"
         },
+        "birthdateRange": {
+            "type": "date_range",
```
steftrenson, line 230: Since ES8 is a lot stricter on these ranges, I wonder if we should also verify the "corectness" of these ranges in the search service as well.

tests/ElasticSearch/JsonDocument/Properties/BirthdateRangeTransformerTest.php
```diff
@@ -0,0 +1,74 @@
...
+        $this->transformer = new BirthdateRangeTransformer();
+    }
+
+    /**
+     * @test
+     */
+    public function it_maps_from_and_to_onto_gte_and_lte(): void
+    {
+        $from = [
+            'birthdateRange' => [
```
steftrenson: I would add 2 extra tests, one with only from and one with only to. These scenarios are possible but now untested.

tests/ElasticSearch/JsonDocument/Properties/BirthdateRangeTransformerTest.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search\ElasticSearch\JsonDocument\Properties;
+
+use PHPUnit\Framework\TestCase;
+
+final class BirthdateRangeTransformerTest extends TestCase
```
steftrenson, line 9: Looking at the whole test, I think this is an ideal candidate to use a [data provider](https://docs.phpunit.de/en/10.5/writing-tests-for-phpunit.html#data-providers). It would reduce the amount of boilerplate in the tests, and allow for quicker expansion of test cases.

src/ElasticSearch/JsonDocument/Properties/BirthdateRangeTransformer.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    public function transform(array $from, array $draft = []): array
+    {
+        if (!isset($from['birthdateRange']) || !is_array($from['birthdateRange'])) {
+            return $draft;
+        }
+
+        $range = $from['birthdateRange'];
+
+        if (!empty($range['from'])) {
```
lucwollants, line 19: We have a rule in our schema that both from and to are always required. Can we rely on this?

src/ElasticSearch/Operations/json/mapping_udb3_core.json
lucwollants: Good remark, not sure we did enforce this on the schema (if possible) or in the HTTP layer code

JonasVHG: We have various tests on correctness when creating/adding birthdate ranges e.g., `features/event/birthdate-range.feature:57` and since there is no historic data with dubious/incorrect data like other, older events, this might not be needed? Still I can add it, if desired.

lucwollants: It seems to be guarded here https://github.com/cultuurnet/udb3-backend/blob/master/src/Model/ValueObject/Audience/BirthdateRange.php and here https://github.com/cultuurnet/apidocs/blob/main/projects/uitdatabank/models/event-birthdateRange.json

lucwollants (approved):

PR #462, opened 2026-05-20
III-6158 Replace DSL lib with own implementation for es8 code path

steftrenson:
### Added
- Add a custom ES8 DSL library (`src/ElasticSearch/DSL/`) to replace `ongr/elasticsearch-dsl` on the ES8 code path, covering all 15 query/sort/aggregation types used in this codebase
- Add `ES8OfferQueryBuilder` and `ES8OrganizerQueryBuilder` backed by the custom DSL, selected at runtime when `elasticsearch.version` is `8`
- Add `NamedBuilderInterface` extending `BuilderInterface` with `getName()`, separating aggregation name from content and making sub-aggregation nesting safe in future
- Add `withShardPreference` to `QueryBuilder`, `OfferQueryBuilderInterface`, and `OrganizerQueryBuilderInterface`
- Add `createUrlParameters` to `QueryBuilder` interface
- Add unit tests and ongr parity tests for every DSL class (30 test files, 94 assertions), explicitly documenting all intentional ES8 divergences
- Add abstract base test classes (`AbstractElasticSearchOfferQueryBuilderTest`, `AbstractElasticSearchOrganizerQueryBuilderTest`) so the full ES5 test suite is also run against the ES8 builders

### Changed
- Wire `ES8OfferQueryBuilder` and `ES8OrganizerQueryBuilder` into the service providers behind the `elasticsearch.version = 8` flag
- Replace `instanceof` checks in `OfferSearchController` and `OrganizerSearchController` with interface method calls
- Make `CardinalityAggregation` accept `field` as a required constructor parameter instead of via `setField()`, eliminating the invalid intermediate state
- Drop `setParameters()` from `FieldSort` — the single call site now passes `nested_path` via the constructor
- Widen `GeoDistanceQuery` `$location` typehint from `object` to `mixed` to match ongr and support string locations
- Add named-key and four-value coordinate formats to `GeoBoundingBoxQuery`
- Add `GT`/`LT` constants and deduplication logic (`gte` takes precedence over `gt`, `lte` over `lt`) to `RangeQuery`
- Add `float` support to `TermQuery` value typehint, matching ongr
- `FieldSort` nested sort now emits ES8 `nested.path` / `nested.filter` structure instead of the legacy `nested_path` flat key
- `GeoShapeQuery::addPreIndexedShape` omits `type` and `relation` fields, which are removed/defaulted in ES8
- `TermsAggregation` and `CardinalityAggregation` now return inner-only content from `toArray()`; `Search` applies the aggregation name externally via `NamedBuilderInterface`

### Fixed
- Fix `BoolQuery::toArray()` returning `[]` for an empty bool clause, which serialises as a JSON array instead of the object `{}` that Elasticsearch expects
- Fix shallow `clone` in `withShardPreference` — replaced with `getClone()` (deep clone via serialize/unserialize) to avoid shared `$search` and `$boolQuery` references between builder copies
- Fix `GeoShapeQuery` silently producing an invalid query (`{"geo_shape":{"":{"indexed_shape":{"id":"","index":"","path":""}}}}`) when `toArray()` is called before `addPreIndexedShape()`

---
Ticket: https://jira.uitdatabank.be/browse/III-6158


lucwollants:
@claude review

steftrenson:
Not being actively worked on, and needs a decision on if we want to go with writing our own DSL library, or opt for a third-party library.

PR #467, opened 2026-05-29
III-7057 Fix recommendation sort

steftrenson:
### Fixed
- Fix `sort[recommendationScore]` returning a misleading 404 "Could not parse query given q parameter as a valid Lucene query" error by replacing `FieldSort::setNestedFilter()` — which serializes the filter under the key `nested` — with `setParameters()` passing `nested_path` and `nested_filter`, which is the correct Elasticsearch 5 sort syntax
- Correct the unit test for `withSortByRecommendationScore` that was asserting the wrong ES sort shape (`nested` instead of `nested_filter`) and therefore passing silently against the broken behaviour

---
Ticket: https://jira.uitdatabank.be/browse/III-7057

### DIFF
--- a/src/ElasticSearch/Offer/ElasticSearchOfferQueryBuilder.php
+++ b/src/ElasticSearch/Offer/ElasticSearchOfferQueryBuilder.php
@@ -545,14 +545,10 @@ final class ElasticSearchOfferQueryBuilder extends AbstractElasticSearchQueryBui
     {
         $fieldSort = new FieldSort('metadata.recommendationFor.score', $sortOrder->toString());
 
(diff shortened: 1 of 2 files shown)


steftrenson:
⚠️ This is not fixed for the ES8 path yet. The behat tests in [udb3-backend](https://github.com/cultuurnet/udb3-backend/pull/2258) will guard this for es8 as well. The fix for the ES8 code path is to use our [custom dsl library](https://github.com/cultuurnet/udb3-search-service/pull/462).

lucwollants (approved):

PR #490, opened 2026-07-03
III-7288 Log document Id

steftrenson:
### Added
- Add `ElasticSearchDocumentCouldNotBeIndexed` exception class in `src/ElasticSearch/` to represent a typed, named failure when ES rejects a document during indexing

### Changed
- `SingleFileIndexationStrategy::indexDocument()` now catches `ElasticsearchException` from the ES client and rethrows it as `ElasticSearchDocumentCouldNotBeIndexed` with the index name and body size in the message, preserving the original exception as `$previous`
- `TransformingJsonDocumentIndexService::index()` now catches `ElasticSearchDocumentCouldNotBeIndexed`, logs it with the document ID, and rethrows it with the document ID in the message — making Sentry issues immediately identifiable without log correlation

---
Ticket: https://jira.uitdatabank.be/browse/III-7288


steftrenson:
@claude pls review

src/JsonDocument/TransformingJsonDocumentIndexService.php
```diff
@@ -49,7 +50,23 @@ public function index(string $documentId, string $documentIri): void
 
         $this->logger->debug("Transformation of {$documentType} {$documentId} finished.");
 
-        $this->searchRepository->save($jsonDocument);
+        try {
+            $this->searchRepository->save($jsonDocument);
+        } catch (ElasticSearchDocumentCouldNotBeIndexed $exception) {
```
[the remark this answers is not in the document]
steftrenson: We'll leave this out of scope for now and rediscuss when it's necessary

src/ElasticSearch/IndexationStrategy/SingleFileIndexationStrategy.php
```diff
@@ -45,7 +47,20 @@ public function indexDocument(
             $params['type'] = $documentType;
         }
 
-        $this->elasticSearchClient->index($params);
+        try {
+            $this->elasticSearchClient->index($params);
+        } catch (ElasticsearchException $e) {
+            throw new ElasticSearchDocumentCouldNotBeIndexed(
```
lucwollants, line 53: Should we not directly log the error here? No need for a new exception and no need for an extra try/catch.

The constructor already takes in a logger.

bertramakers: @steftrenson can you clarify whether it's best to throw and log it in another class (and why), or would logging it here also be okay? If the latter it might be nice to avoid the extra code for the custom exception and the catch statement.

steftrenson:
@bertramakers @lucwollants You're right, this implementation was too convoluted. I've made a simpler version https://github.com/cultuurnet/udb3-search-service/pull/502/changes

PR #493, opened 2026-07-06
III-7215 Add hasChildcare offer search filter

grubolsch:
## What

Adds a new boolean search parameter `hasChildcare` on the offer search endpoint that lets API consumers filter offers by the presence of childcare hours.

- `hasChildcare=true` → only offers that have at least one sub-event or opening hour with a `childcare` range configured.
- `hasChildcare=false` → only offers that have no `childcare` configured on any sub-event or opening hour.
- Omitted → behaviour unchanged (no childcare filtering).

## Ticket
https://jira.publiq.be/browse/III-7215

## Related
https://github.com/cultuurnet/udb3-backend/pull/2314


grubolsch:
@claude

src/ElasticSearch/JsonDocument/Properties/CalendarTransformer.php
lucwollants, line 1: This file seems to have a lot of "overhead". Very verbose comments and methods like `determineHasChildcare` make it harder to read than required.

bertramakers:
This functionality happened to come up in an unrelated meeting (about the calendar summary). Corneel asked whether this filter would always match on `?childCare=true` if at least one subEvent has one subEvent with childcare hours, or if it would take the date range you're filtering on into account and only return events that have a subEvent  within the period that you're searching with childcare hours.

For example, scenario 1:
- Event A has 2 subevents, one on 2026-01-01 with childcare until 18:00, and one on 2026-01-08 without childcare
- I look for events on no specific date, with childcare => Event A matches
- I look for events happening on 2026-01-01 with childcare => Event A matches
- I look for events happening on 2026-01-08 with childcare => Event  A also matches while it technically doesn't have childcare on that day

While in scenario 2:
- I look for events on no specific date, with childcare => Event A matches
- I look for events happening on 2026-01-01 with childcare => Event A matches
- I look for events happening on 2026-01-08 with childcare => Event A doesn't match because it doesn't have childcare on that day

If I understand correctly this PRs implements scenario 1, but the PMs prefer scenario 2 if it's technically possible.

It just so happens that we already have this kind of behavior for other subEvent properties like `status` and `bookingAvailability`. On those filters we check if there is also a `dateRange` query parameter set or not, and if so we look at the `status` or `bookingAvailability` of the specific subEvents that match with that dateRange. If no dateRange is set, we look at the top-level `status` or `bookingAvailability`. (Comparable with the overall "childcare" in this PR)

So I think we can keep the top-level "childcare" and filter on that if no `dateRange` is set, but ideally we would additionally also take `dateRange` into account when set. You can see how that is implemented already for `status` and `bookingAvailability` here and extend it to also work the same for `childCare`: https://github.com/cultuurnet/udb3-search-service/blob/0b673b29a4f76c24da299618a15c46e7bfb883bf/src/Http/Offer/RequestParser/CalendarOfferRequestParser.php#L42-L79

grubolsch:
I did the change regarding taking into account start/end dates in a search

src/ElasticSearch/Offer/ElasticSearchOfferQueryBuilder.php
```diff
@@ -245,6 +245,8 @@ public function withSubEventFilter(SubEventQueryParameters $subEventQueryParamet
             $this->guardNaturalIntegerRange('localTime', new Time($localTimeFrom), new Time($localTimeTo));
         }
 
+        $hasChildcare = $subEventQueryParameters->getHasChildcare();
```
bertramakers: Nit: I'd move this up so it's together with the other "$subEventQueryParameters->get...()" calls in this method

src/ElasticSearch/Offer/ElasticSearchOfferQueryBuilder.php
```diff
@@ -302,6 +304,10 @@ public function withSubEventFilter(SubEventQueryParameters $subEventQueryParamet
             $queries[] = new MatchQuery('subEvent.bookingAvailability', $bookingAvailability);
         }
 
+        if ($hasChildcare !== null) {
+            $queries[] = new TermQuery('subEvent.hasChildcare', $hasChildcare);
```
bertramakers, line 307: Since it's been some time since I worked on this I was wondering whether `TermQuery` is correct to query a boolean field, since there is also a boolean filter, but after looking it up in the docs this is indeed correct 👍

src/ElasticSearch/Operations/SchemaVersions.php
```diff
@@ -6,6 +6,6 @@
 
 final class SchemaVersions
 {
-    public const UDB3_CORE = 20260701173900;
+    public const UDB3_CORE = 20260714120000;
```
bertramakers, line 9: Good that you remembered! (Until we simplify this)

tests/Http/OfferSearchControllerTest.php
```diff
@@ -732,6 +734,58 @@ public function it_converts_the_uitpas_toggle_parameter_to_a_correct_boolean(
         $this->controller->__invoke(new ApiRequest($request));
     }
 
+    /**
+     * @test
+     * @param bool|string|int $stringValue
+     * @dataProvider booleanStringDataProvider
+     */
+    public function it_converts_the_has_childcare_toggle_parameter_to_a_correct_boolean(
```
bertramakers, line 742: Always good to add a test like this for boolean query parameters because I remember the conversion from string to bool is more tricky than you'd imagine 😄 👍

tests/ElasticSearch/JsonDocument/data/place/indexed-with-regions.json
```diff
@@ -139,7 +141,10 @@
     },
     "geo": {
         "type": "Point",
-        "coordinates": [4.70373, 50.8876694]
+        "coordinates": [
+            4.70373,
+            50.8876694
+        ]
```
bertramakers, line 147: Nit: I'm a bit confused why there are some minor unrelated changes like this one, that don't seem to serve any purpose (especially here with the coordinates). Claude? 😄

bertramakers (approved):
Looks good! Consistent with the same approach for other properties. Good test coverage 👍 Just 2 small nits, the other comments is just me pointing out things I appreciate.

docs/III-7215-childcare-feature-tests-prompt.md
```diff
@@ -0,0 +1,174 @@
+# Prompt: write feature (Behat) tests for the `hasChildcare` search filter
```
bertramakers: Was this committed on purpose or accidentally?

src/ElasticSearch/Operations/SchemaVersions.php
grubolsch: I added it to the claude.md rules. :_D

PR #495, opened 2026-07-08
III-7216 overnight-filter: add hasOvernight offer search filter

grubolsch:
## What

Adds a boolean `hasOvernight` search parameter to the offer search endpoint so consumers can include or exclude offers that involve an overnight stay.

- `hasOvernight=true` → only offers with **at least one** sub-event flagged `overnight: true`.
- `hasOvernight=false` → only offers with **no** overnight sub-event.
- **Omitted** → behaviour unchanged (no overnight filtering).

**Rollup (event level):** if at least one sub-event has `overnight: true`, the whole event is considered overnight, so a partial overnight event matches `hasOvernight=true`.

## Ticket
[III-7216](https://jira.publiq.be/browse/III-7216)


grubolsch:
@claude

grubolsch:
Jenkins run: https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/817/

docs/III-7216-overnight-feature-tests-prompt.md
lucwollants, line 1: Seems like an unwanted temporary file?

lucwollants (commented):
The documentation https://github.com/cultuurnet/apidocs/pull/557/changes states that adding `dateFrom` and `dateTo` takes into account the `hasOvernight` from the sub-events, but this pull request only works at the top level?

In the description, there is no mention of this or a follow-up pull request, so I assume it is missing?

grubolsch:
overNight is now also on subEvents so you can use the dateRange on it.

lucwollants:
> overNight is now also on subEvents so you can use the dateRange on it.

It's indeed added, but it seems that a `true` value is never covered in the unit tests?

grubolsch:
I see 2 test with an overnight=true combined with a date filter:
- tests/ElasticSearch/Offer/ElasticSearchOfferQueryBuilderTest.php:977 (test het bouwen van SubEventQueryParameters)
-  tests/Http/OfferSearchControllerTest.php:846 (http laag)

And of course there is the feature tests as well.

grubolsch:
Now I read the code again, I might misunderstand you, my tests are doing searches with hasOvernight=true, but they don't contain events with it. This seems in line with the others tests in `OfferSearchControllerTest`.
Maybe you want me to write a test with actual subevents with hasOvernight, but I cannot find examples how to mock this, without doing actual ES searches. Given that I am already testing it in a feature test, is that enough?

lucwollants:
> I see 2 test with an overnight=true combined with a date filter:
> 
> * tests/ElasticSearch/Offer/ElasticSearchOfferQueryBuilderTest.php:977 (test het bouwen van SubEventQueryParameters)
> * tests/Http/OfferSearchControllerTest.php:846 (http laag)
> 
> And of course there is the feature tests as well.

I should have been more precise. It always seems to be false in the JSON event body, and the `EventTransformerTest` only covers the `false` case.

grubolsch:
I now understand: https://github.com/cultuurnet/udb3-search-service/pull/495/changes/916fb0eea515a289e847f91f3438c10e290383d8

docs/calendar-indexing.md
```diff
@@ -86,10 +86,10 @@ The JSON-LD shape is the same as for events.
...
 Not every field in the JSON-LD ends up in Elasticsearch. Some fields are **read at index time and then discarded**. Others are **stored and queryable**.
 
-| Role | Fields | Stored in ES? |
-|---|---|---|
-| Source only | `calendarType`, `startDate`, `endDate`, `openingHours`, `childcare` | No, consumed to build the indexed fields below |
-| Indexed | `dateRange`, `localTimeRange`, `subEvent[]`, `availableRange`, `hasChildcare` | Yes, queryable |
+| Role | Fields                                                                           | Stored in ES? |
+|---|----------------------------------------------------------------------------------|---|
+| Source only | `calendarType`, `startDate`, `endDate`, `openingHours`, `overnight`, `childcare` | No, consumed to build the indexed fields below |
+| Indexed | `dateRange`, `localTimeRange`, `subEvent[]`, `availableRange`, `hasOvernight`, `childcare    | Yes, queryable |
```
lucwollants, line 92: Should this be `hasChildcare` as the indexed field? Because I think we convert the childcare object to just false/true in the index?

lucwollants (approved):

PR #496, opened 2026-07-09
III-7276 Cap subEvent count to stay under the ES nested-object limit

steftrenson:
### Added
- Add `SubEventCapTransformer`, wired immediately after `CalendarTransformer` in `OfferTransformer`, which truncates an offer's `subEvent` array to a configurable maximum and logs a warning identifying the offer id and both counts when truncation happens, preventing Elasticsearch from rejecting documents that exceed `index.mapping.nested_objects.limit`
- Add unit tests for `SubEventCapTransformer` covering missing/under/at/over-cap scenarios, and integration tests in `EventTransformerTest`/`PlaceTransformerTest` that run a real calendar through the full transformer pipeline to guard against `SubEventCapTransformer` ever being reordered before `CalendarTransformer`
- Document the `subEvent` cap and its config key in `docs/calendar-indexing.md`

### Changed
- Add a required `int $subEventCap` constructor parameter to `OfferTransformer`, `EventTransformer`, and `PlaceTransformer` to thread the configurable cap down to `SubEventCapTransformer`

---
Ticket: https://jira.uitdatabank.be/browse/III-7276

---
appconfig PR: https://github.com/cultuurnet/appconfig/pull/1417

### DIFF
--- a/app/Event/EventIndexationServiceProvider.php
+++ b/app/Event/EventIndexationServiceProvider.php
@@ -34,7 +34,8 @@ final class EventIndexationServiceProvider extends BaseServiceProvider
                             $this->get('logger.amqp.udb3')
                         ),
                         new PathEndIdUrlParser(),
-                        $this->get(GeoShapeQueryRegionService::class)
+                        $this->get(GeoShapeQueryRegionService::class),
+                        (int) ($this->parameter('elasticsearch.sub_event_cap') ?? 9900)
                     )
                 );
 
--- a/app/Event/EventIndexationServiceProvider.php
+++ b/app/Event/EventIndexationServiceProvider.php
@@ -6,6 +6,7 @@ namespace CultuurNet\UDB3\SearchService\Event;
 
 use CultuurNet\UDB3\Search\ElasticSearch\ElasticSearchDocumentRepository;
 use CultuurNet\UDB3\Search\ElasticSearch\JsonDocument\EventTransformer;
+use CultuurNet\UDB3\Search\ElasticSearch\JsonDocument\Properties\SubEventCapTransformer;
 use CultuurNet\UDB3\Search\ElasticSearch\PathEndIdUrlParser;
 use CultuurNet\UDB3\Search\ElasticSearch\Region\GeoShapeQueryRegionService;
 use CultuurNet\UDB3\Search\Event\EventSearchProjector;
@@ -35,7 +36,7 @@ final class EventIndexationServiceProvider extends BaseServiceProvider
                         ),
                         new PathEndIdUrlParser(),
                         $this->get(GeoShapeQueryRegionService::class),
-                        (int) ($this->parameter('elasticsearch.sub_event_cap') ?? 9900)
+                        (int) ($this->parameter('elasticsearch.sub_event_cap') ?? SubEventCapTransformer::DEFAULT_CAP)
                     )
                 );
 
--- a/app/Event/EventIndexationServiceProvider.php
+++ b/app/Event/EventIndexationServiceProvider.php
@@ -6,7 +6,6 @@ namespace CultuurNet\UDB3\SearchService\Event;
 
 use CultuurNet\UDB3\Search\ElasticSearch\ElasticSearchDocumentRepository;
 use CultuurNet\UDB3\Search\ElasticSearch\JsonDocument\EventTransformer;
-use CultuurNet\UDB3\Search\ElasticSearch\JsonDocument\Properties\SubEventCapTransformer;
 use CultuurNet\UDB3\Search\ElasticSearch\PathEndIdUrlParser;
 use CultuurNet\UDB3\Search\ElasticSearch\Region\GeoShapeQueryRegionService;
 use CultuurNet\UDB3\Search\Event\EventSearchProjector;
@@ -35,8 +34,7 @@ final class EventIndexationServiceProvider extends BaseServiceProvider
                             $this->get('logger.amqp.udb3')
                         ),
                         new PathEndIdUrlParser(),
-                        $this->get(GeoShapeQueryRegionService::class),
-                        (int) ($this->parameter('elasticsearch.sub_event_cap') ?? SubEventCapTransformer::DEFAULT_CAP)
+                        $this->get(GeoShapeQueryRegionService::class)
                     )
                 );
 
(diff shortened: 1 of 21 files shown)


steftrenson:
@claude pls review

app/Event/EventIndexationServiceProvider.php
```diff
@@ -34,7 +34,8 @@ function (): EventSearchProjector {
                             $this->get('logger.amqp.udb3')
                         ),
                         new PathEndIdUrlParser(),
-                        $this->get(GeoShapeQueryRegionService::class)
+                        $this->get(GeoShapeQueryRegionService::class),
+                        (int) ($this->parameter('elasticsearch.sub_event_cap') ?? 9900)
```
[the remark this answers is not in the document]
steftrenson: Updated the description and linked to the appconfig pr.

src/ElasticSearch/JsonDocument/Properties/SubEventCapTransformer.php
```diff
@@ -0,0 +1,60 @@
...
+     */
+    private function truncate(array $subEvents, string $id): array
+    {
+        $originalCount = count($subEvents);
+
+        $this->logger->logWarning(
+            "subEvent truncated from {$originalCount} to {$this->maxSubEvents} entries for {$id}."
+        );
+
+        return array_slice($subEvents, 0, $this->maxSubEvents);
```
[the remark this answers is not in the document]
steftrenson: This is indeed a conscious decision to keep this as simple as possible for now.

src/ElasticSearch/JsonDocument/Properties/SubEventCapTransformer.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        if (!isset($draft['subEvent']) || !is_array($draft['subEvent'])) {
+            return $draft;
+        }
+
+        if (count($draft['subEvent']) <= $this->subEventCap) {
+            return $draft;
+        }
+
+        $id = $from['@id'] ?? 'unknown';
```
grubolsch: Can be inline

src/ElasticSearch/JsonDocument/Properties/SubEventCapTransformer.php
```diff
@@ -0,0 +1,62 @@
...
+    public const DEFAULT_CAP = 9900;
+
+    private JsonTransformerLogger $logger;
+
+    private int $subEventCap;
+
+    public function __construct(JsonTransformerLogger $logger, int $subEventCap)
+    {
+        $this->logger = $logger;
+        $this->subEventCap = $subEventCap;
```
bertramakers, line 21: Nit: To me the constant is sufficient for now and it doesn't need to be configurable. It's the same for all environments, and even the same as the constant right now. I don't really see a need to ever change this to a value that's different per environment. And if we ever need to change it globally we can update the constant. Also because the constructor argument needs to be passed through multiple classes it seems. But if you think it's best to also have it in the config I don't mind

src/ElasticSearch/JsonDocument/Properties/SubEventCapTransformer.php
```diff
@@ -0,0 +1,62 @@
...
+     * This is a naive, non-negotiated cut-off: no attempt is made to pick which entries matter most (e.g.
+     * biasing towards future-relevant slots, or sampling across the full range). For a periodic calendar
+     * with a very long startDate-endDate range and dense opening hours, this can drop several years from
+     * the tail end of the range. That's accepted: the scenario is rare, and losing a few years off an
+     * already multi-decade range is a small relative loss compared to the complexity of a smarter
+     * selection strategy.
+     *
+     * @param array[] $subEvents
+     */
+    private function truncate(array $subEvents, string $id): array
```
grubolsch, line 52: I would avoid those @params that don't add any extra type information.
I would avoid having such large docblocks, I feel this explains more that needs to be said about this short code.
Maybe summerize it like: "This is a simple cut-off strategy. For very large periodic calendars with dense opening hours, the end of the date range may be truncated, and could use years of information, but these will lay far in the future."

src/ElasticSearch/JsonDocument/OfferTransformer.php
```diff
@@ -53,6 +55,8 @@ public function __construct(
             new NameTransformer($logger),
             new DescriptionTransformer(),
             new CalendarTransformer($logger),
+            // Must run after CalendarTransformer — caps draft['subEvent'] which CalendarTransformer writes.
+            new SubEventCapTransformer($logger, $subEventCap),
```
bertramakers, line 59: Could we keep it a separate class but inject/call it inside CalendarTransformer to avoid the risk that it runs in the incorrect order? The comment helps of course but if we can mitigate it completely that would be even better

grubolsch (approved):

grubolsch:
Minor remarks, already approved

src/ElasticSearch/JsonDocument/OfferTransformer.php
bertramakers: Okay I now see the unit test that should prevent this. That also works of course, but keeping my comment up for your consideration

bertramakers (approved):
Thanks @steftrenson , I have two smaller comments for your consideration but nothing blocking to me so already approving. Let me know if you agree or not with my suggestions and if you'd like me to review any new commits (if any)

src/ElasticSearch/JsonDocument/Properties/SubEventCapTransformer.php
steftrenson: Ok, I have an alternate approach, because I want it to at least be configurable from the outside, because testing is a nightmare otherwise.

src/ElasticSearch/JsonDocument/Properties/SubEventCapTransformer.php
steftrenson: I don't fully agree with this. It cleanly documents why this approach was chosen, and what the trade-offs are. This gives a future dev (probably with an llm) sufficient context on why this approach was chosen. I'll remove the docblock though.

src/ElasticSearch/JsonDocument/OfferTransformer.php
steftrenson: Conceptually I don't think the calendarTransformer should know about the fact that events are being capped. The calendarTransformer already has quite some responsibilities, and I wouldn't add this responsibility into the mix. The tests should be sufficient to guarantee the order doesn't get switched.

bertramakers (approved):

PR #504, opened 2026-07-17
III-7190 matching birthdate range resultset fixes

JonasVHG:
Added

  - New top-level matchingBirthdateRanges field on offer search responses: when the query contains a birthdate range (via advanced q birthdateRange:[from TO to]
  clauses — flat, grouped or repeated — or the structured birthdateRangeFrom/birthdateRangeTo params), the response includes one entry per queried range with its
  from, to and the list of returned event @ids that qualify for it. Omitted when no birthdate range is queried.
  - BirthdateRangeQueryStringParser: single home for the birthdateRange: query-string patterns; parses a q string into BirthdateRange[].
  - MatchingBirthdateRangesResolver: resolves the queried ranges from a request and computes per-range matches, mirroring createBirthdateRangeQuery (indexed
  birthdateRange intersects the queried dates, or typicalAgeRange intersects the equivalent age range excluding all-ages events).
  - ElasticSearchOfferQueryBuilder::withBirthdateRangeMatchFields(): widens _source with birthdateRange/typicalAgeRange/allAges, applied only when a birthdate
  range is queried.

  Changed

  - OfferSearchController resolves queried birthdate ranges after request parsing (widening _source when needed) and attaches matchingBirthdateRanges to the
  response next to the existing facet output; wired via OfferSearchControllerFactory.
  - BirthdateRangeToTypicalAgeRangeQueryStringFactory now reuses the shared BirthdateRangeQueryStringParser patterns instead of its own private regex constants
  (behaviour unchanged).

  Removed
  
  - Duplicate birthdateRange: regex constants from BirthdateRangeToTypicalAgeRangeQueryStringFactory (moved into BirthdateRangeQueryStringParser).

  Related PR

  - https://github.com/cultuurnet/udb3-backend/pull/2325

  ---
  Ticket: https://jira.publiq.be/browse/III-7190

### DIFF
--- a/src/Http/Offer/MatchingBirthdateRangesResolver.php
+++ b/src/Http/Offer/MatchingBirthdateRangesResolver.php
@@ -1,0 +1,190 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CultuurNet\UDB3\Search\Http\Offer;
+
+use Cake\Chronos\Chronos;
+use CultuurNet\UDB3\Search\ElasticSearch\BirthdateRangeQueryStringParser;
+use CultuurNet\UDB3\Search\Http\ApiRequestInterface;
+use CultuurNet\UDB3\Search\Offer\BirthdateRange;
+use CultuurNet\UDB3\Search\ReadModel\JsonDocument;
+use CultuurNet\UDB3\Search\UnsupportedParameterValue;
+use DateTimeImmutable;
+use stdClass;
+
+/**
+ * Resolves the `matchingBirthdateRanges` response field.
+ *
+ * A request "queries" one or more birthdate ranges, either through the advanced query
+ * (`q=birthdateRange:[from TO to]`, possibly grouped or repeated) or through the structured
+ * `birthdateRangeFrom`/`birthdateRangeTo` parameters. For every queried range this resolver reports
+ * which of the returned events qualify for it, so a consumer can show per-child "geschikt voor"
+ * hints when multiple children are searched in one query.
+ *
+ * The match test mirrors ElasticSearchOfferQueryBuilder::createBirthdateRangeQuery(): an event
+ * qualifies when its indexed `birthdateRange` intersects the queried date range, or when its
+ * `typicalAgeRange` intersects the equivalent age range (excluding "all ages" events). The
+ * birthdate <-> age conversion is relative to "now", matching the query-time behaviour everywhere
+ * else in the codebase.
+ */
+final class MatchingBirthdateRangesResolver
+{
+    private BirthdateRangeQueryStringParser $queryStringParser;
+
+    private DateTimeImmutable $now;
+
+    public function __construct(BirthdateRangeQueryStringParser $queryStringParser, ?DateTimeImmutable $now = null)
+    {
+        $this->queryStringParser = $queryStringParser;
+        $this->now = $now ?? new Chronos();
+    }
+
+    /**
+     * The birthdate ranges expressed by the request, in order and without duplicates.
+     *
+     * @return BirthdateRange[]
+     */
+    public function queriedRanges(ApiRequestInterface $request): array
+    {
+        $ranges = [];
+
+        if ($request->hasQueryParam('q')) {
+            $ranges = $this->queryStringParser->parse((string) $request->getQueryParam('q'), $this->now);
+        }
+
+        $structured = $this->structuredRange($request);
+        if ($structured !== null) {
+            $ranges[] = $structured;
+        }
+
+        return $this->deduplicate($ranges);
+    }
+
+    /**
+     * @param BirthdateRange[] $queriedRanges
+     * @param JsonDocument[] $results
+     * @return array<int, array{from: string, to: string, matches: string[]}>
+     */
+    public function match(array $queriedRanges, array $results): array
+    {
+        $documents = array_map(static fn (JsonDocument $result): stdClass => $result->getBody(), $results);
+
+        return array_map(
+            function (BirthdateRange $range) use ($documents): array {
+                $matches = [];
+                foreach ($documents as $document) {
+                    if ($this->documentMatchesRange($document, $range)) {
+                        $matches[] = (string) ($document->{'@id'} ?? '');
+                    }
+                }
+
+                return [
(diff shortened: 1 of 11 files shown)


JonasVHG:
@claude review.

src/Http/Offer/MatchingBirthdateRangesResolver.php
```diff
@@ -0,0 +1,190 @@
...
+        } catch (UnsupportedParameterValue $e) {
+            return null;
+        }
+    }
+
+    /**
+     * @param BirthdateRange[] $ranges
+     * @return BirthdateRange[]
+     */
+    private function deduplicate(array $ranges): array
```
grubolsch, line 106: Just food for thought, we should maybe look if we can port the collection classes we have in UDB3 Backend (CultuurNet\UDB3\Model\ValueObject\Collection) to SAPI3, so we can jsut pass collection objects, that would make this code more readable.
You don't need to change this now

tests/Http/OfferSearchControllerTest.php
```diff
@@ -380,6 +392,76 @@ public function it_returns_a_paged_collection_of_search_results_based_on_request
         $this->assertEquals($expectedJsonResponse, $actualJsonResponse);
     }
 
+    /**
+     * @test
+     */
+    public function it_adds_matching_birthdate_ranges_when_a_birthdate_range_is_queried(): void
```
grubolsch, line 398: Maybe we should add a check that matchingBirthdateRanges is missing when you do not query a birthdate range. Probably already implicit tested somewhere else but good to make a formal contract about it.

tests/ElasticSearch/BirthdateRangeToTypicalAgeRangeQueryStringFactoryTest.php
```diff
@@ -67,13 +67,93 @@ public function it_leaves_a_query_without_a_birthdate_range_unchanged(): void
...
+            . ' AND name.nl:foo'
+        );
+
+        $this->assertEquals($expected, $actual);
+    }
+
+    /**
+     * @test
+     */
+    public function it_skips_an_invalid_range_inside_a_grouped_birthdate_range(): void
```
grubolsch, line 135: Good test, but should we also have a test where every subrange is invalid? There is a specific if in the code for this edge case that could be tested:

                if ($ageClauses === []) {
                    return $original;
                }

src/Http/Offer/MatchingBirthdateRangesResolver.php
```diff
@@ -0,0 +1,190 @@
...
+ * which of the returned events qualify for it, so a consumer can show per-child "geschikt voor"
+ * hints when multiple children are searched in one query.
+ *
+ * The match test mirrors ElasticSearchOfferQueryBuilder::createBirthdateRangeQuery(): an event
+ * qualifies when its indexed `birthdateRange` intersects the queried date range, or when its
+ * `typicalAgeRange` intersects the equivalent age range (excluding "all ages" events). The
+ * birthdate <-> age conversion is relative to "now", matching the query-time behaviour everywhere
+ * else in the codebase.
+ */
+final class MatchingBirthdateRangesResolver
```
grubolsch, line 24: I think this comment is very verbose, a lot of this can just be read from the code, focus more on the why?
Here is a suggestion how to shorten it:

 * Resolves which queried birthdate ranges match each returned event.
 *
 * Supports both `q=birthdateRangeFrom:[from TO to]` and
 * `birthdateRangeFrom`/`birthdateRangeTo` parameters. An event matches if its
 * `birthdateRange` or (non-"all ages") `typicalAgeRange` intersects the
 * queried range, using the same "now"-relative logic as the search query.
 */

grubolsch:
If I understand correctly `documentMatchesRange()` and
`createBirthdateRangeQuery()` kinda do the same thing, but they are on
different levels (post-fetch PHP check on already-returned documents vs.
the actual ES filter query that decides the result set in the first
place). 
My only problem is that they have to keep matching to the same
result set, or in the future we risk changing the boundaries in one of
the two, but not the other, right now the only thing keeping them in
sync is a doc-comment ("mirrors ..."), nothing fails if one changes
without the other. Since this drives a public API field, a silent drift
here would be a hard-to-notice correctness bug (e.g. `matchingBirthdateRanges`
claiming a match for an event that isn't even in the result set, or vice
versa).

Can we maybe make a test that starts with the same fixture and checks if they return the same boundries?

grubolsch (changes requested):

src/Http/Offer/MatchingBirthdateRangesResolver.php
```diff
@@ -0,0 +1,190 @@
...
+        $to = isset($birthdateRange->lte) ? DateTimeImmutable::createFromFormat('!Y-m-d', (string) $birthdateRange->lte) : null;
+        $from = $from instanceof DateTimeImmutable ? $from : null;
+        $to = $to instanceof DateTimeImmutable ? $to : null;
+
+        if ($from === null && $to === null) {
+            return false;
+        }
+
+        // Intersection with [range->from, range->to]; a missing bound is unbounded.
+        if ($from !== null && $from > $range->getTo()) {
```
grubolsch, line 144: Comment does not add much

tests/ElasticSearch/BirthdateRangeToTypicalAgeRangeQueryStringFactoryTest.php
JonasVHG: Done in previous PR.

src/Http/Offer/MatchingBirthdateRangesResolver.php
JonasVHG: Indeed, worth creating a ticket for it or mentioning it in a weekly.

grubolsch: Mentioned it in the biweekly notes.

grubolsch:
I run the jenkins tests and then I will merge.

https://jenkins.publiq.be/job/uitdatabank-acceptance-tests/823/

grubolsch:
@claude

grubolsch (approved):

PR #508, opened 2026-07-23
III-7357 Support multiple comma-separated ranges in birthdateRangeFrom/birthdateRangeTo

grubolsch:
## Summary
- Adds comma-separated, positionally-paired multi-range support to `birthdateRangeFrom`/`birthdateRangeTo`, e.g. `?birthdateRangeFrom=2024-01-01,2026-01-01&birthdateRangeTo=2024-12-30,2026-12-30`, giving the structured params OR-multiple-ranges parity with what `q=birthdateRange:([...] OR [...])` already supports.

## Related
https://github.com/cultuurnet/udb3-backend/pull/2329

## Ticket
https://jira.publiq.be/browse/III-7357

### DIFF
--- a/src/Http/Parameters/ArrayParameterBagAdapter.php
+++ b/src/Http/Parameters/ArrayParameterBagAdapter.php
@@ -171,6 +171,29 @@ final class ArrayParameterBagAdapter implements ParameterBagInterface
     }
 
     /**
+     * @return DateTimeImmutable[]
+     */
+    public function getExplodedDateFromParameter(
+        string $parameterName,
+        ?string $defaultValueAsString = null,
+        string $delimiter = ','
+    ): array {
+        $callback = function ($asString) use ($parameterName): DateTimeImmutable {
+            $date = $this->parseDate((string) $asString);
+
+            if ($date === null) {
+                throw new UnsupportedParameterValue(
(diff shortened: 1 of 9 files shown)


grubolsch:
@claude

src/Http/Parameters/ArrayParameterBagAdapter.php
```diff
@@ -228,6 +208,20 @@ private function ensureCallback(callable $callback = null): callable
         return $passThroughCallback;
     }
 
+    private function parseDateOrFail(string $value, string $parameterName): DateTimeImmutable
+    {
+        // Trim so comma-separated values with spaces, e.g. "2020-01-01, 2022-06-15", parse as expected.
+        $date = $this->parseDate(trim($value));
```
bertramakers: Nit: Since `parseDate` is only used inside this method and nowhere else, I would suggest to instead refactor `parseDate` to always throw instead of returning `null` when the parsing fails. Throwing an exception is usually safer anyway, so I don't think we need the option of parsing a date without it throwing if it fails. And even then, it could be handled with a try/catch.

bertramakers (approved):
Looks good, just one small nit but approving either way since it doesn't change much about the current solution, only some internals that could be simplified for futureproofing.

PR #516, opened 2026-08-03
III-7347 Remove the matchingBirthdateRanges response field

lucwollants:
## Summary
- Removes the `matchingBirthdateRanges` field from the offer search response, returning it to its original shape. It was never documented, and the PR that would have documented it (cultuurnet/apidocs#561) was closed.
- Deletes `MatchingBirthdateRangesResolver`, `ElasticSearchOfferQueryBuilder::withBirthdateRangeMatchFields()` and the controller wiring.
- Deletes `BirthdateRangeQueryStringParser`. It only existed to share its two regex patterns between the resolver and `BirthdateRangeToTypicalAgeRangeQueryStringFactory`; with the resolver gone the factory is the only consumer left, so the patterns move back into it as `private const` — which is where they lived before #504.

## Not affected
Querying birthdate ranges is unchanged. Both `birthdateRangeFrom`/`birthdateRangeTo` and the `q=birthdateRange:[...]` form keep working, including multiple OR-ed ranges via `q`, and the `typicalAgeRange` fallback is untouched on both paths.

No mapping JSON changed, so no `SchemaVersions` bump and no reindex — the only Elasticsearch-side change is that `_source` is no longer widened at query time.

## Notes for review
`BirthdateRangeMatchingConsistencyTest` is deleted rather than partially kept. Its stated purpose was "the ES filter and the resolver agree on the same document"; with the resolver gone, the remaining half is a hand-rolled PHP simulation of Elasticsearch range semantics checked against fixtures, which can only catch a wrong bounds calculation, not a wrong query. Query shape stays covered by `ElasticSearchOfferQueryBuilderTest` (three `withBirthdateRangeFilter` cases, including multi-range), and real boundary behaviour by the acceptance tests against a live Elasticsearch. Happy to revive the boundary fixtures as a plain query-builder test if reviewers prefer.

## Related
- https://github.com/cultuurnet/udb3-backend/pull/2335

## Ticket
https://jira.publiq.be/browse/III-7347


lucwollants:
Superseded by #517 — reopened under the correct ticket (III-7190, the ticket that added `matchingBirthdateRanges`). Closed automatically by a branch rename, not intentionally.

PR #521, opened 2026-08-05
Index both the age and birthdate range on events

lucwollants:
### Added

Every event now carries both `typicalAgeRange` and `birthdateRange` in Elasticsearch. Whichever of the two was not filled in is derived from the other at index time, using the event start date as reference. Permanent events use the indexing time instead.

This lets `q=typicalAgeRange:[...]` and `q=birthdateRange:[...]` return the same events no matter which of the two was entered, without rewriting the query at search time.

### How

- `TypicalAgeRangeTransformer` indexes `typicalAgeRange` as a gte/lte range and sets the `allAges` flag.
- A new `AgeTransformer` derives the missing range from the one that is present. It runs after the two source transformers.
- The core schema version is bumped so the derived ranges get backfilled onto existing documents on reindex.

### All-ages events

An all-ages event suits every birthdate, so it gets an unbounded `birthdateRange` with both bounds `null`. It then matches every birthdate query, the same way its `typicalAgeRange` already matches every age query. Anyone who does not want these events excludes them with `allAges=false`. This settles the all-ages question raised on the review of apidocs #563.

### Not in this PR

- Removing the query-time conversion, which becomes redundant once both fields are indexed. That is a follow-up.
- The `typicalAgeRangeConverted` and `birthdateRangeConverted` response fields.

Ticket: https://jira.publiq.be/browse/III-7367


src/ElasticSearch/JsonDocument/Properties/AgeTransformer.php
```diff
@@ -0,0 +1,156 @@
...
+        $startDate = $from['startDate'] ?? null;
+
+        if (is_string($startDate)) {
+            $parsed = DateTimeImmutable::createFromFormat(DateTime::ATOM, $startDate);
+
+            return $parsed === false ? null : self::fromDateString($parsed->format('Y-m-d'));
+        }
+
+        if (($from['calendarType'] ?? null) === 'permanent') {
+            return self::fromDateString($this->now->format('Y-m-d'));
```
steftrenson: It feels a bit weird that a dateTimeImmutable is formatted to a string, passed to a private function only to return a new dateTimeImmutable.

steftrenson (approved):

src/ElasticSearch/JsonDocument/Properties/AgeTransformer.php
lucwollants: Made it more explicit in 9a3103e and also used `Chronos` to handle lead years inside bb4fe08

src/ElasticSearch/JsonDocument/Properties/AgeTransformer.php
```diff
@@ -0,0 +1,154 @@
...
+        $draft['birthdateRange'] = [
+            'gte' => $oldestBirthdate,
+            'lte' => $youngestBirthdate,
+        ];
+
+        return $draft;
+    }
+
+    /**
+     * A ChronosDate holds no time and no offset, so a start date at 00:30+01:00 counts as that day
```
bertramakers, line 137: Is there a real scenario in which this can happen? The `birthdateRange` from and to values are always just a date right? This is just theoretical safety? Or are there real life scenarios in which this can be a problem?

bertramakers (approved):

src/ElasticSearch/JsonDocument/Properties/AgeTransformer.php
lucwollants: The method `toPlainDate` is also used for the `startDate` and `now` when the calendar type is `permanent`. In those cases, the stripping of the hours is needed. So, to keep the code cleaner, I also used it to create the dates from the `gte` and `lte` values from `birthdateRange`.

This also puts the try/catch around the `ChronosDate` constructor in one code location.

PR #522, opened 2026-08-06
Expose the converted age and birthdate ranges in search results

lucwollants:
### Added

Search results now expose `typicalAgeRangeConverted` and `birthdateRangeConverted`. When a range was derived at index time, because the editor entered the other one, it is returned under a "Converted" name so a consumer can tell it apart from an entered value.

- A new `ConvertedAgesJsonTransformer` runs in the embedded result pipeline. It reads the indexed range and adds the converted field only when the original JSON-LD did not carry it.
- An unbounded range, from an open or all ages age range, has no from/to pair, so its converted birthdate range is left out. The `typicalAgeRange` still conveys it to the consumer.

Ticket: https://jira.publiq.be/browse/III-7367


src/ElasticSearch/JsonDocument/ConvertedAgesJsonTransformer.php
```diff
@@ -0,0 +1,50 @@
...
+namespace CultuurNet\UDB3\Search\ElasticSearch\JsonDocument;
+
+use CultuurNet\UDB3\Search\JsonDocument\JsonTransformer;
+
+/**
+ * Adds typicalAgeRangeConverted and birthdateRangeConverted to a search result. A range that is
+ * missing from the original JSON-LD but present on the indexed document was derived at index time,
+ * so it is exposed under a "Converted" name to keep it apart from a value the editor entered.
+ */
+final class ConvertedAgesJsonTransformer implements JsonTransformer
```
steftrenson, line 13: Personally I think this belongs to the AgeTransformer, rather than in a separate transformer. This one needs to run after that AgeTransformer has ran, or it will give wrong results.

lucwollants: This class `ConvertedAgesJsonTransformer` is wired inside the `ResultTransformerFactory` and is used for displaying the results. The `AgeTransformer` is wired inside the `EventTransformer` used at index time.

bertramakers (approved):

